PR #21, opened 2015-01-29
JANDEX-29: Jandex fails to reindex index JAR file

arcivanov:
Do not copy existing jandex.idx to the output zip


arcivanov:
Would it be also possible to produce 1.2.3.Final to be able to use the fixed jandex utility with WildFly 8.2?
The 2.x Alpha currently published has v6 of Jandex not compatible with WFLY8.2

n1hility:
Sure thanks for the PR

PR #24, opened 2015-07-20
[JANDEX-33] Fixes an infinite loop when a JAR contains a CRC error on a .class file

iweiss:


n1hility:
Thanks for the patch. I went with a slightly different version though since 0 isn't necessarily EOF

PR #27, opened 2016-01-06
Version information has been changed.

maozturk:
Version should has been changed since final version has been released.


n1hility:
Thanks

PR #28, opened 2016-02-25
JANDEX-35 Failure to resolve method type variable when using class-level bounded wildcard generics

kurobako:
https://issues.jboss.org/browse/JANDEX-35
fixes failure to resolve method type variable when using class-level bounded wildcard generics

signature list is now sorted before an attempt to resolve type variables is made; class signatures are processed first


n1hility:
Thanks for the PR, i took your test (preserving your commit authorship) and added a slightly more efficient variant that doesn't require resorting and additional memory allocation (#29)

PR #30, opened 2016-05-16
Add basic test for nested classes.

tremes:


n1hility:
Thanks for the patch. Although I ended up going with something that preserves compatibility and also corrects the value from flags() 

Superseded by #31

PR #36, opened 2017-12-05
JANDEX-41 Just ignoring module-info.class files but processing other…

gunnarmorling:
… class files with level 53

This is not adding support for module info descriptors, but just applying Scott's previous work-around a bit more selectively: only module-info.class files will be ignored, whereas regular class files will be processed, also if they have been compiled to the Java 9 class file format. IIUC, the 53 version doesn't affect any class files other than module-info.class files.


n1hility:
Thanks!

PR #41, opened 2018-11-15
Update unit tests to cover Comparator properties as well

Sanne:
Implementation looks great, I just added a couple more tests to improve confidence on the `compareTo`.


n1hility:
Danke!

PR #43, opened 2018-11-30
Method param names

FroMage:
Support method parameter names:

- Bumped index version
- Read method param name attribute
- Default to debug info if missing
- Align method param name indices to method param annotation indices (not types)

This is pretty much required for RESTEasy extensions indexed by MP-OpenAPI


FroMage:
I think it'd be good to run the test suite a second time, using javac 1.8 with the `-parameters` flag to test this in Maven (I've only tested it in the IDE), but I don't know how to do that in Maven.

n1hility:
Excellent work! Thank you!

Issue #45, opened 2018-12-13
Support a reliable way to get a parameter type


gsmet:
Currently, to get the parameter type, we need to do something like:
```
methodParameterInfo.method().parameters().get(methodParameterInfo.position())
```

The issue with this is that `parameters()` includes the synthetic parameters (confirmed by @FroMage ) whereas they are not counted in the result of `position()`, leading to incorrect results if we have synthetic parameters.

@FroMage implemented a nice user friendly API to get the parameter name. We should do the same for the parameter type.

FroMage:
For backwards compat I can't change the behaviour of `parameters()`, but I can find another solution.

FroMage:
Well, actually, parameter annotations will tell us the number of non-synthetic params. Parameter names (if present in the bytecode) will also tell us. But if you method has no annotation and no parameter name it has no info on synthetic params. And also, javac and ECJ generate different synthetic params for enums. So it's not that easy.

FroMage:
```java
    /**
     * Returns the nesting type of this class, which could either be a standard top level class, an inner class,
     * an anonymous class, or a local class.
     *
     * <p>For historical reasons, static nested classes are returned as <code>INNER</code>. You can differentiate
     * between a non-static nested class (inner class) and a static nested class by calling
     * {@link java.lang.reflect.Modifier#isStatic(int)} on the return of {@link #flags()} </p>
     *
     * @return the nesting type of this class
     */
    public NestingType nestingType() {
...
```

This is lying at least with ECJ, the `ACC_STATIC` flag is missing in the bytecode.

FroMage:
Actually it works and `javap` is just hiding it.

FroMage:
OK, so:

- I can get it to work for toplevel and inner classes. 
- For local classes it requires an `Index` parameter to look up the declaring method's class to see if it's static. 
- For anonymous classes it doesn't look like Jandex actually supports them (I can't find any instance of an anonymous class that responds anything else than TOPLEVEL).
- For enums I can't figure out how to make it work. ECJ produces the following bytecode:

```
  private org.jboss.jandex.test.BasicTestCase$Enum(java.lang.String, int, byte);
    descriptor: (Ljava/lang/String;IB)V
```

Where the `String, int` are synthetic and the byte is real, but javac produces this:

```
  private org.jboss.jandex.test.BasicTestCase$Enum(byte);
    descriptor: (Ljava/lang/String;IB)V
```

[...]

PR #48, opened 2019-01-02
Synthetic param types

FroMage:
Support a reliable way to get a parameter type for #45

This throws if you try it on a local/anonymous type, but that's OK because you will never care about these constructors since you can't call them with anything useful.

### DIFF
--- a/src/main/java/org/jboss/jandex/MethodInfo.java
+++ b/src/main/java/org/jboss/jandex/MethodInfo.java
@@ -141,14 +141,46 @@ public final class MethodInfo implements AnnotationTarget {
     }
 
     /**
-     * Returns the name of the given parameter.
+     * Returns the name of the given parameter, omitting synthetic parameters.
      * @param i the parameter index
-     * @return the name of the given parameter, or null.
+     * @return the name of the given parameter, omitting synthetic parameters, or null.
      */
     public final String parameterName(int i) {
         return methodInternal.parameterName(i);
     }
+
+    /**
+     * Returns the type of the given parameter, omitting synthetic parameters. This
+     * does not support local classes.
+     * @param i the parameter index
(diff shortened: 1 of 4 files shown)


src/main/java/org/jboss/jandex/MethodInfo.java
```diff
@@ -141,14 +141,46 @@ public final String name() {
...
     }
+
+    /**
+     * Returns the type of the given parameter, omitting synthetic parameters. This
+     * does not support local classes.
+     * @param i the parameter index
+     * @return the type of the given parameter, omitting synthetic parameters, or null.
+     * @throws UnsupportedOperationException if this method's container class is a local class or an enum.
+     */
+    public Type realParameter(int parameter) {
```
gsmet: Not totally excited by the name of the method, given we have `parameterName()` just above, also omitting synthetic parameters.

Other than that, it definitely looks like what I was looking for when I asked for it.

FroMage: `parameterTypeSkippingSyntheticParams()`?

MikeEdgar: How about an overloaded method simply called `parameterType` to allow default behavior of using only "natural" parameters and also a way to access the synthetic type without direct access to the collection?

```java
public Type parameterType(int parameter) {
    return parameterType(parameter, false);
}

public Type parameterType(int parameter, boolean allowSynthetic) {
    // Same logic from realParameter in PR, but considering allowSynthetic value
}
```

Ladicek:
Note to self: need to take into account that synthetic constructs come in 2 shapes, _mandated_ (JLS says they must be present) and others.

Ladicek:
Superseded by #192.

PR #52, opened 2019-02-18
JANDEX-49: Top-level ClassInfo read from an index file now reports the correct NestingType

gastaldi:
@n1hility please review


mkouba:
I think we can close this PR as @n1hility introduced a more complex fix in https://github.com/wildfly/jandex/commit/ce4dd9a57e4ac6f308da57aac01fcede1a94370a. @gastaldi could you pls test your app with Jason's fix?

PR #59, opened 2019-04-09
Set Automatic-Module-Name and fix maven-enforcer-plugin problem

ljnelson:
This PR bumps the version of the maven-enforcer-plugin so that it can cope with the version strings of modern JDKs and won't exit with an `ExceptionInInitializerError` when invoked.  It also sets an `Automatic-Module-Name` in the `META-INF/MANIFEST.MF` file.  I followed the module naming guidelines referenced by Stephen Coleburne here: https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html

This is obviously a stopgap measure until this library can be truly modularized but it at least "claims" the module name with no compatibility issues.

Signed-off-by: Laird Nelson <ljnelson@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,6 +27,15 @@
     </licenses>
 
     <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-enforcer-plugin</artifactId>
+                    <version>3.0.0-M2</version>
+                </plugin>
+            </plugins>
+        </pluginManagement>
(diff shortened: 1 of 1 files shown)


n1hility:
Thanks for the PR. I think this is reasonable.

FWIW I think it's unlikely that Jandex will ever utilize Java modules beyond this. It supports pre Java 9 vms (as far back as Java 6), and for Java 9+ we also wish to support common classpath usage. While you can add a modules descriptor and let it be ignored, that also means that any security or isolation is also ignored, so it's best to just rely on package security.

Issue #58, opened 2019-04-08
Indexer option to control ClassInfo sorting


neon-dev:
[SmallRye's OpenAPI implementation](https://github.com/smallrye/smallrye-open-api/) uses Jandex. Not only for annotation processing but also to get general reflective class info. Unfortunately object fields are sorted by Jandex: https://github.com/wildfly/jandex/blob/61d1115e0a0f15dcfb2b0bbd082100f847963529/src/main/java/org/jboss/jandex/ClassInfo.java#L557
This ultimately alters the field order of all objects appearing in the openapi file, though in most cases declaration order is desired.

It would therefore be great to have an [Indexer](https://github.com/wildfly/jandex/blob/master/src/main/java/org/jboss/jandex/Indexer.java) option to configure ClassInfo's field sorting behavior. Maybe for completeness also for [methods](https://github.com/wildfly/jandex/blob/61d1115e0a0f15dcfb2b0bbd082100f847963529/src/main/java/org/jboss/jandex/ClassInfo.java#L581).

Thanks!

neon-dev:
@n1hility any update about this topic? If it's out of scope (even though a simple flag :smirk:) it would be nice to know. In that case the linked issue will be more complex to fix, but at least we have clarity about it.

vlsi:
@neon-dev , can you please clarify what do you mean by "declaration order"?
Java specification has no guarantees that the fields and methods will appear in exactly the same order in the bytecode.
Then, `Class#getMethods` has the following javadoc:

>The elements in the returned array are not sorted and are not in any particular order.

Sorting entries might help build reproducibility (see https://reproducible-builds.org/ ).
Then "storing original method positions" would make "index merging" slightly harder.

neon-dev:
@vlsi correct, but in practice I've not yet seen the compiler shift methods and fields around. So it'd be great to have a way to tell Jandex to retain the original/input order, even if it may not be consistent over different bytecode versions. Meaning whatever `getDeclaredMethods` and `getDeclaredFields` returns, take it as is.

PR #61, opened 2019-05-24
Add control over field and method sorting 

neon-dev:
PR for issue #58.

### DIFF
--- a/src/main/java/org/jboss/jandex/ClassInfo.java
+++ b/src/main/java/org/jboss/jandex/ClassInfo.java
@@ -57,6 +57,8 @@ public final class ClassInfo implements AnnotationTarget {
     private Type[] typeParameters;
     private MethodInternal[] methods;
     private FieldInternal[] fields;
+    private boolean areMethodsSorted;
+    private boolean areFieldsSorted;
     private boolean hasNoArgsConstructor;
     private NestingInfo nestingInfo;
 
@@ -321,8 +323,8 @@ public final class ClassInfo implements AnnotationTarget {
      */
     public final MethodInfo method(String name, Type... parameters) {
         MethodInternal key = new MethodInternal(Utils.toUTF8(name), MethodInternal.EMPTY_PARAMETER_NAMES, parameters, null, (short) 0);
-        int i = Arrays.binarySearch(methods, key, MethodInternal.NAME_AND_PARAMETER_COMPONENT_COMPARATOR);
-        return i >= 0 ? new MethodInfo(this, methods[i]) : null;
+        MethodInternal method = findMethod(key, true);
+        return method == null ? null : new MethodInfo(this, method);
     }
 
     /**
@@ -336,13 +338,25 @@ public final class ClassInfo implements AnnotationTarget {
      */
     public final MethodInfo firstMethod(String name) {
         MethodInternal key = new MethodInternal(Utils.toUTF8(name), MethodInternal.EMPTY_PARAMETER_NAMES, Type.EMPTY_ARRAY, null, (short) 0);
-        int i = Arrays.binarySearch(methods, key, MethodInternal.NAME_AND_PARAMETER_COMPONENT_COMPARATOR);
-        if (i < -methods.length) {
-            return null;
-        }
+        MethodInternal method = findMethod(key, false);
+        return method == null ? null : new MethodInfo(this, method);
+    }
 
-        MethodInfo method = new MethodInfo(this,i >= 0 ? methods[i] : methods[++i * -1]);
-        return method.name().equals(name) ? method : null;
+    private MethodInternal findMethod(MethodInternal key, boolean exactMatch) {
+        if (areMethodsSorted) {
+            int i = Arrays.binarySearch(methods, key, MethodInternal.NAME_AND_PARAMETER_COMPONENT_COMPARATOR);
+            if (i >= 0 || !exactMatch && i >= -methods.length && methods[i = -(i + 1)].name().equals(key.name())) {
+                return methods[i];
+            }
+        } else {
+            for (MethodInternal method : methods) {
+                int diff = MethodInternal.NAME_AND_PARAMETER_COMPONENT_COMPARATOR.compare(method, key);
+                if (diff == 0 || !exactMatch && method.name().equals(key.name())) {
+                    return method;
+                }
+            }
+        }
+        return null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kenfinnigan:
That's probably easier.

What do you think @neon-dev?

neon-dev:
If it doesn't cause any issues elsewhere, sure why not. Never looked into why fields get sorted in the first place. Since Jandex is heavily used during Thorntail startup for example I assumed there was a good reason.

Method sorting though is something I am not that sure about. For one it's handy that your OpenAPI implementation takes advantage of this, but on the other hand developers could sort methods in the classes themselves. The question is if most would actually (want to) do that.
I guess if it can be removed from Jandex it'd be worth considering adding method sorting in https://github.com/smallrye/smallrye-open-api/
This would also enable tag scoped method sorting (relevant when using the same tag on different classes).

dmlloyd:
The main concern I would have would be that they are sorted in order to facilitate index merging or binary searches or something like that.  @n1hility do you remember if this is the case?

neon-dev:
Nvm my comment about method sorting in regards to OpenAPI. What we care about there are paths, not method names, so that's what needs sorting. I just confused them since I mostly try to name them equally.

neon-dev:
Fun fact, method binary searches use the comparator that is also used for initial sorting: https://github.com/wildfly/jandex/search?q=binarySearch+filename%3AClassInfo.java&unscoped_q=binarySearch+filename%3AClassInfo.java
So the actual array order doesn't matter, except maybe slightly in performance terms, if at all.

n1hility:
> The main concern I would have would be that they are sorted in order to facilitate index merging or binary searches or something like that. @n1hility do you remember if this is the case?

Right @dmlloyd that's the problem. The sorting is for binary searches.

n1hility:
> Fun fact, method binary searches use the comparator that is also used for initial sorting: https://github.com/wildfly/jandex/search?q=binarySearch+filename%3AClassInfo.java&unscoped_q=binarySearch+filename%3AClassInfo.java
> So the actual array order doesn't matter, except maybe slightly in performance terms, if at all.

@neon-dev I don't follow what you mean. Binary searches can only function when the array is sorted according to the search algorithm (in this case retrieving a field by it's name)

The reason it uses binary searches over an additional allocated hash map was to keep memory usage limited as that was a strong original goal. I could add declaration ordering info but it would mean increasing memory usage by (field + method count) * machine word + object header + padding.

I was originally skeptical this was of much value because there is no hard requirement that a class file actually follow declaration order from the original java source (in fact the JLS explicitly allows reordering). If however it can be shown that declarative order is in practice reliable (would require researching bytecode output of various versions and implementations of java) then I could rethink this.

neon-dev:
@n1hility my bad. Until now I thought the binary search would search in the order the comparator specifies, no matter the actual ordering. Should've read the doc before talking about a function I've never used before myself :no_mouth:

n1hility:
@neon-dev > @n1hility my bad. Until now I thought the binary search would search in the order the comparator specifies, no matter the actual ordering. Should've read the doc before talking about a function I've never used before myself 😶

Ah ok, gotcha. Sorry for such a long delay getting back to you, for some reason I missed the notifications in while travelling. Just to be clear from my earlier reply I am totally fine with adding this extra info. I just want to double check that the reliability aspect so the extra memory usage is worth it. Have you looked into this (ordering mechanics of javac, ecj etc from java6 and onward)?

neon-dev:
Oh, no problem. I have to thank for taking the time, especially since this is not of importance for Jandex itself.
So you think it'd be feasable to add declaration ordering info? I'm not experienced on the compiler side of things, so sorry, can't tell you anything helpful here.
Since I also don't know much about Jandex other than that it's used to start up a wildfly instance, let me ask a silly question:
Would it be too slow just iterating these arrays to find the element? I assume the method and field arrays generally are relatively small in size, so my uneducated guess (hope) is, that this wouldn't be a noticeable difference. Or did I entirely miss something else here? Not that I'm again talking weird stuff because I'm assuming the wrong things.

neon-dev:
Since field and method search was broken in my PR when disabling sorting, I added a fallback logic. Maybe this also helps you testing performance differences.

dmlloyd:
Maybe we don't want a commit called 'Whoopsie' in there ;)

neon-dev:
Was tempted to rename it to "sorry", but that wouldn't have helped much I guess :D

neon-dev:
I just rebased everything again into a single commit for a cleaner history and to fix a typo.

n1hility:
Hi @neon-dev thank you for your continued contribution on this. I have been toying with the notion of a heuristics based approach, storing/indexing in class file occurrence order but limiting it to cases under 256 members and only employing bsearch at > 4 elements.  Draft implementation is here: 
https://github.com/n1hility/jandex/commit/3108a5a9158b36acd2d1b0ed75fae5662de6097e

This results in about 220KB extra memory required for indexing rt.jar (out of around 26 or so megs). There is some potential savings possible through some other small optimizations. The search performance is important since its a very common pattern to reference / lookup names, and its common to have pretty large member sets. 

A rough breakdown:
```
Fields
------
0-7       : 18748
8-15      : 1082
16-255    : 958
256-65535 : 6

Methods
--------
0-7       : 15019
8-15      : 2850
16-255    : 2918
256-65535 : 7
```

neon-dev:
Just checked your commit and I must say that's quite the effort and an interesting approach. I never expected to cause that much work with this issue/PR.
In case you didn't already consider it, here are a few ideas regarding further optimizations:
Your statistics suggests it could be worth checking if member searching performs any worse in the real world when setting MAX_LINEAR to 5-7 to save another bit of memory and overhead.
Also you could init the sort indexes on demand for the same reason. So refactoring everything to call fieldSortArray() / methodSortArray() instead of accessing the fields directly. One thing to keep in mind then would be thread safety for sure.

Thanks for your time and work.

neon-dev:
Hey. Did you run further tests in the last weeks? I'm curious how different settings with your solution work out. Also interested how much worse my PR would perform in comparison. And of course I'd like to see this PR's work come to fruition because of the referenced issue.

MikeEdgar:
@n1hility , @neon-dev - I spent a little bit of time looking at this and considered a slightly different approach. What are your thoughts on storing the original positions (up to 256 of each field and member positions) and simply using them only for an explicit request for an unsorted collection? This would leave the existing sorting and searching intact and also limit the memory usage to that used in Jason's commit.

Please see [commit 701951f](https://github.com/MikeEdgar/jandex/commit/701951f1c4e397fc5c8ab604cc38d0599fc3bb42).

neon-dev:
@MikeEdgar sorry for the late reply. I still stand by my very first reply in this issue and will answer this from a user's perspective, since I don't know the technical requirements to Jandex. So from a pure caller's perspective unsorted views should be the standard and always available, while sorted ones may or may not be available. Anyone can sort collections however he likes, but unsorting is impossible.

As has been discussed very early on, the main goal of sorting was to accelerate lookups, not to actually get sorted results. I'd only sort if really necessary for said performance improvements while keeping all of this internal and retaining support for unsorted views of such elements.
Set up some benchmarks and some real world scenarios and test if there's a noticeable difference in startup time / throughput / etc. between each version.
Also want to note that there are no methods prefixed with `get`, so a caller should not expect optimal performance in the first place and code with this in mind (i.e. avoid retrieving the same element over and over, implement caching, etc.). Without the need for sorting this issue would be trivial. So maybe make sure if it really is necessary performance wise? Also constantly sorting while indexing surely also takes some time which could be saved without it. Just an idea, as I said I have no clue about who uses Jandex and what the performance impact would be. That's what's really interesting to find out!

Having said this, I'd be fine with any implementation you guys decide. Thanks for all your work and ideas!

neon-dev:
Thanks for the great work in #73!

PR #63, opened 2019-06-06
Add isEnumConstant implementation to FieldInfo

### LINKED ISSUE #62: Add isEnumConstant to FieldInfo
`java.lang.reflect.Field` has the method `isEnumConstant` that uses the package private value `Modifier.ENUM`. The value of the constant `0x4000` is defined in the class file format to indicate that a field is "declared as an element of an enum" (https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.5-200-A.1).

It would be a nice convenience to implement a similar method on `FieldInfo` checking for `0x4000` in the field's flags.

```java
package org.jboss.jandex;
...
public final class FieldInfo implements AnnotationTarget {
    ...
    static final int ENUM = 0x00004000;
    ....
    public boolean isEnumConstant() {
        return (flags() & ENUM) != 0;
    }
    ....
}
```

This is closely related (it seems) to #54 and #60.

MikeEdgar:
Fixes #62 

* include basic documentation
* include basic unit test
* remove unused imports in FieldInfo

### DIFF
--- a/src/main/java/org/jboss/jandex/FieldInfo.java
+++ b/src/main/java/org/jboss/jandex/FieldInfo.java
@@ -18,11 +18,7 @@
 
 package org.jboss.jandex;
 
-import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Modifier;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


MikeEdgar:
Please provide feedback on this change.

MikeEdgar:
Please provide feedback on this change.

n1hility:
Thanks for the contribution! 

I think it makes since but we should add methods for all of the other types. I'm happy to do that if you don't have time for it.

I'll slate this for the next minor release.

mkouba:
@n1hility This PR can be closed in favor of https://github.com/wildfly/jandex/pull/76.

PR #64, opened 2019-07-12
Add type interface specialization for MethodInfo and FieldInfo

geoand:
This can be useful when iterating over annotations of methods and fields
for example when one wants to collect and/or manipulate the list of
annotations that class "properties" have

### DIFF
--- a/src/main/java/org/jboss/jandex/ContainingAnnotationTarget.java
+++ b/src/main/java/org/jboss/jandex/ContainingAnnotationTarget.java
@@ -1,0 +1,16 @@
+package org.jboss.jandex;
+
+import java.util.List;
+
+/**
+ * Holds methods common {@link org.jboss.jandex.FieldInfo} and {@link org.jboss.jandex.MethodInfo}.
+ *
+ */
+public interface ContainingAnnotationTarget extends AnnotationTarget {
+
+    List<AnnotationInstance> annotations();
+
+    AnnotationInstance annotation(DotName name);
+
+    boolean hasAnnotation(DotName name);
+}
--- a/src/main/java/org/jboss/jandex/FieldInfo.java
+++ b/src/main/java/org/jboss/jandex/FieldInfo.java
@@ -34,7 +34,7 @@ import java.util.List;
  * @author Jason T. Greene
  *
  */
-public final class FieldInfo implements AnnotationTarget {
+public final class FieldInfo implements ContainingAnnotationTarget {
     private ClassInfo clazz;
     private FieldInternal internal;
 
@@ -107,6 +107,7 @@ public final class FieldInfo implements AnnotationTarget {
      *
      * @return the list of annotations on this field
      */
+    @Override
     public List<AnnotationInstance> annotations() {
         return internal.annotations();
     }
@@ -118,6 +119,7 @@ public final class FieldInfo implements AnnotationTarget {
      * @param name the name of the annotation to locate on this field
      * @return the annotation if found, otherwise, null
      */
+    @Override
     public final AnnotationInstance annotation(DotName name) {
         return  internal.annotation(name);
     }
@@ -130,6 +132,7 @@ public final class FieldInfo implements AnnotationTarget {
      * @param name the name of the annotation to look for
      * @return true if the annotation is present, false otherwise
      */
+    @Override
     public final boolean hasAnnotation(DotName name) {
         return internal.hasAnnotation(name);
     }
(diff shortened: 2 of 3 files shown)


geoand:
The name of the interface I added is most likely pretty bad but I couldn't think of a better one

MikeEdgar:
@geoand , @n1hility - would it make sense to have these methods (or similarly named ones to avoid conflicts) directly on `AnnotationTarget`? It seems it would be very useful to generically access annotations directly on a target (and nested annotations when applicable) via the common interface itself.

geoand:
Can show an example of the user experience you have in mind?

MikeEdgar:
It would become useful in situations where a common piece of code exists to process a particular annotation that may appear on a class, method, or field. Something similar to the interface `java.lang.reflect.AnnotatedElement`, but without the inherited annotations of course.

Granted, it's easy enough to switch on the `target.kind()` and call the related `as` method on the target to get the concrete type, but that ends up being "utility" code that might be better located in Jandex. We have several cases of this in SmallRye OpenAPI.

This method would be able to process a particular method on any type of target without itself needing to check the `target`'s type in the process. 

```java
void processMyAnnotation(AnnotationTarget target) {
    DotName myAnnoName = DotName.createSimple("org.example.MyAnnotation");

    if (target.hasAnnotation(myAnnoName)) {
        AnnotationInstance instance = target.annotation(myAnnoName);
        // ... processing logic
    }
}
```

geoand:
That does seem reasonable.

@Ladicek @FroMage what is your take?

MikeEdgar:
In fact, it may even make sense to have a higher level interface with those methods so that `Type` can implement it as well. One thing that also might be useful would be to limit the returned annotations to those directly on the target. E.g. method annotations would not include the parameters. That, in conjunction with the changes in PR #75 would be a nice addition to the API in my opinion.

Ladicek:
If we could have these methods directly on `AnnotationTarget`, that would be great. After all, if I have something that is "target of annotations", why can't I get those annotations?

Another option, if retriveing annotations from type usages is not exactly straightforward (which it isn't in Jandex) and not exactly important (which it isn't in my experience), would be to introduce a concept of _declaration_. If you look at the Jandex `AnnotationTarget`, you will find that there are 2 kinds of annotation targets: declarations and types. Yet there's no interface that would correspond to declarations -- you have to deal with classes, fields, methods and method parameters separately. If we could create an interface, say, `Declaration` or `DeclarationInfo` or `AnnotatedDeclaration`, which would lie between `{Class,Field,Method,MethodParameter}Info` and `AnnotationTarget`, that would work too IMHO.

Something like (very close to this PR):

```java
interface DeclarationInfo {
    Collection<AnnotationInstance> annotations();
    AnnotationInstance annotation(DotName name);
    Collection<AnnotationInstance> annotationsWithRepeatable(DotName name, IndexView index);
    boolean hasAnnotation(DotName name);
}
```

This PR implements this for `MethodInfo` and `FieldInfo`. Implementing it for `MethodParameterInfo` would be rather straightforward. Unfortunately, `ClassInfo` already has a method `annotations` which has a different signature :-(

FroMage:
> If we could have these methods directly on AnnotationTarget, that would be great. After all, if I have something that is "target of annotations", why can't I get those annotations?

Absolutely. Otherwise we need to rename `AnnotationTarget` to `BlackholeAnnotationTarget` since we can't get anything out of it.

MikeEdgar:
> This PR implements this for `MethodInfo` and `FieldInfo`. Implementing it for `MethodParameterInfo` would be rather straightforward. Unfortunately, `ClassInfo` already has a method `annotations` which has a different signature :-(

I suggest that they would be new method names to support an opt-in approach, particularly if the response changes (e.g. annotations on methods shouldn't return everything, just items directly on the method). This would clear up ambiguities such as what is the expected return from `MethodInfo.annotation(DotName)` when the same annotation appears on the method as well as on multiple parameters.

Ladicek:
For some reason, I got to this discussion again, and got an idea. Similarly to how `java.lang.Class` has `getMethods` and `getDeclaredMethods`, we could have `annotations` and `declaredAnnotations`. The former being current behavior (annotations _contained_ in that element, that is, present on the element or elements nested inside), the latter being the new behavior (annotations present only on the element).

So, we could have these 4 methods:

```java
Collection<AnnotationInstance> declaredAnnotations();
AnnotationInstance declaredAnnotation(DotName name);
Collection<AnnotationInstance> declaredAnnotationsWithRepeatable(DotName name, IndexView index);
boolean hasDeclaredAnnotation(DotName name);
```

It should be possible to add these methods directly to `AnnotationTarget`, but if that sounds wrong to anyone, the `DeclarationInfo` idea would still make sense IMHO.

geoand:
Both sound fine to me

PR #70, opened 2019-12-23
Fix AnnotationValue methods names and Javadoc 

gwenneg:
I noticed this typo while using this method from a Quarkus extension.

I'll let you judge whether or not this is reasonnable change considering it is a breaking change for anyone who's using the old method name.

### DIFF
--- a/src/main/java/org/jboss/jandex/AnnotationValue.java
+++ b/src/main/java/org/jboss/jandex/AnnotationValue.java
@@ -119,11 +119,27 @@ public abstract class AnnotationValue {
         return new FloatValue(name, f);
     }
 
+    /**
+     * @deprecated Use {@link #createDoubleValue(String, double)} instead.
+     */
+    @Deprecated
     public static AnnotationValue createDouleValue(String name, double d) {
+        return createDoubleValue(name, d);
+    }
+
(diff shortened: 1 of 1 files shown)


src/main/java/org/jboss/jandex/AnnotationValue.java
```diff
@@ -123,7 +123,7 @@ public static AnnotationValue createDouleValue(String name, double d) {
         return new DoubleValue(name, d);
     }
 
-    public static AnnotationValue createLongalue(String name, long l) {
```
gsmet: There is another typo in the method above `createDouleValue`.

gwenneg: Good catch! Fixed. I also corrected two Javadoc typos.

I guess I should make the old methods deprecated and add the corrected signatures to make this fix non-breaking. WDYT?

gsmet: Yes, I think that would be safer.

gwenneg: The old methods are back and deprecated, this is now a non-breaking PR :)

n1hility:
Thanks!

PR #72, opened 2020-02-03
Fix #71 Reproducible index lookups

### LINKED ISSUE #71: Reproducible index lookups
Calling the same sequence of Jandex methods for the same set of JARs should yeld the same results (incl. the ordering of entries). 

A PR follows.

ppalaga:


src/main/java/org/jboss/jandex/FieldInfo.java
```diff
@@ -34,7 +34,7 @@
  * @author Jason T. Greene
  *
  */
-public final class FieldInfo implements AnnotationTarget {
+public final class FieldInfo implements AnnotationTarget, Comparable<FieldInfo> {
```
gsmet: I'm surprised you don't have the equivalent for `MethodInfo`?

ppalaga: `MethodInfo` is made `Comparable` too in this PR. See below. Or you mean something else?

gsmet: Ah no, silly me, I missed it. Sorry for the noise!

n1hility:
See comment in #71

gsmet:
@Ladicek I think we should revive this PR as it's going to be one of the first element of the journey to make the Quarkus build reproducible.

Ladicek:
The really funny thing is that Jandex has actually been producing reproducible output since version 3.1.0 (see #286). Quarkus has tried to rely on it when it comes to build caching, which uncovered a bug, which was fixed in 3.1.1 (see #308). Since then, I haven't heard from the build caching folks about reproducibility problems.

Fortunately, most of the changes here were not necessary, and I really wouldn't like to apply them.

gsmet:
OK, perfect then, thanks!

PR #73, opened 2020-02-16
Store original member positions for unsorted iteration

### LINKED ISSUE #58: Indexer option to control ClassInfo sorting

MikeEdgar:
Fixes #58 
Supersedes #61 

This change supports the storage of class member positions, as originally scanned from the byte stream. The positions are only utilized when client code requests a list of `unsortedFields` or `unsortedMethods`.

Indexing the rt.jar file from JRE 1.8.0_232 produces the following index size:
1. 4,836,637 bytes using release version 2.1.2.Final of Jandex
2. 5,237,838 bytes with the changes in this PR

A very simple test to load the two indexes shows that the version in this PR uses approximately 1,200k more heap memory (28,345,736 `main` thread Bytes) than 2.1.2.Final (27,128,488 `main` thread Bytes)

```java
public class Load {
    public static void main(String[] args) throws Exception {
        IndexReader reader = new IndexReader(new FileInputStream(args[0]));
        Index index = reader.read();
        Thread.sleep(600000L);
    }
}
```

### DIFF
--- a/src/test/java/org/jboss/jandex/test/ClassInfoMemberPositionTestCase.java
+++ b/src/test/java/org/jboss/jandex/test/ClassInfoMemberPositionTestCase.java
@@ -1,0 +1,679 @@
+/*
+ * JBoss, Home of Professional Open Source.
+ * Copyright 2020 Red Hat, Inc., and individual contributors
+ * as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jboss.jandex.test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.List;
+
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.Index;
+import org.jboss.jandex.IndexReader;
+import org.jboss.jandex.IndexWriter;
+import org.jboss.jandex.Indexer;
+import org.jboss.jandex.MethodInfo;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ClassInfoMemberPositionTestCase {
+
+    private Index index;
+
+    @Before
+    public void setUp() throws IOException {
+        Indexer indexer = new Indexer();
+        String prefix = "org/jboss/jandex/test/ClassInfoMemberPositionTestCase$";
+        indexer.index(getClass().getClassLoader().getResourceAsStream(prefix + "TestEntity.class"));
+        indexer.index(getClass().getClassLoader().getResourceAsStream(prefix + "MaxSizeTestEntity.class"));
+        indexer.index(getClass().getClassLoader().getResourceAsStream(prefix + "OverMaxSizeTestEntity.class"));
+        this.index = indexer.complete();
+    }
+
+    @Test
+    public void testMembersUnsorted() {
+        assertOriginalPositions(index);
+    }
+
+    @Test
+    public void testMembersUnsortedWithFile() throws IOException {
+        File tmpIndex = File.createTempFile("cls", ".idx");
+        OutputStream output = null;
+        try {
+            output = new FileOutputStream(tmpIndex);
+            IndexWriter writer = new IndexWriter(output);
+            writer.write(this.index);
+        } finally {
+            if (output != null) {
+                output.close();
+            }
+        }
+
+        InputStream input = null;
+        Index fileIndex;
+        try {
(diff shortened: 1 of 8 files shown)


n1hility:
Hey @MikeEdgar , Sorry I haven't been able to get to this yet. I'll take a look next week. 

It's hard to tell for sure from your numbers since we might have been using different versions of rt.jars, but the comparison with my earlier prototype looks like it might use around 4x the memory usage. This is probably due to the prototype not storing the sorted values for sizes <=4 (binary searching offers negligible improvements with small lists).

When I have talked to other major users of jandex, there was some hesitancy around the trade-off since true order use cases are less common but everyone takes the cost, that said I lean towards adding it in a future minor rev.

MikeEdgar:
Re-based the branch on current master to keep current. Please let me know if any changes should be made, etc.

CC: @n1hility

MikeEdgar:
@n1hility - I took another look at this and optimized it a bit further by using the byte pool provided by `NameTable`. With this change the memory increase is reduced to about 370k (from 1,200k in earlier commits of this PR). Testing was against the OpenJDK `1.8.0_272` rt.jar comparing against version `2.2.1.Final`

Memory used by `2.2.1.Final` in `main` thread: 27,281,992 B
Memory used by this PR in `main` thread: 27,651,480 B

Please take another look and let me know if this is any more tolerable :-D

MikeEdgar:
Latest commit includes the removal of several unnecessary object allocations (empty arrays and eager `Map` wrapping) to further reduce memory requirements.

Memory used by `2.2.1.Final` in `main` thread: 27,281,992 B
Memory used by this PR in `main` thread: 25,976,208 B

This results in a net *reduction* of around 1.3 M.

vlsi:
> This results in a net reduction of around 1.3 M.

Can the improvement be separated into its own PR?
Currently, it is hard to judge the impact of this PR.

MikeEdgar:
Hi @vlsi - it could. Those changes are limited to commit 9fdd8dd. Do you mean it's hard to judge the impact of the PR on memory usage or something else?

The "baseline" was OpenJDK `1.8.0_272`'s rt.jar loaded into memory with version `2.2.1.Final`. In that test, the main thread consumed 27,281,992 B of memory. With the changes in this PR, the memory consumed drops to 25,976,208 B. Without the functional changes from the earlier commits, the memory consumed would be approx. 370k less. That 370k is for the `byte` arrays holding the class member positions as read from the original class files.

The majority of the savings comes from avoiding the allocation of empty `HashMap` instances in `IndexReaderV2`. The information below shows the delta with the PR (captured via VisualVM). Rows with a delta less than 1k were omitted.

Name | Live Bytes | Live Objects
-- | -- | --
byte[] | 323816 | 9137
char[] | 6344 | 220
int[] | -7312 | -35
java.io.ObjectStreamClass | -1976 | -19
java.io.ObjectStreamClass$WeakClassKey | -10336 | -323
java.lang.Class | 1152 | 11
java.lang.Class[] | 12696 | 545
java.lang.Class$ReflectionData | 9240 | 165
java.lang.invoke.MemberName | 1288 | 23
java.lang.Long | -2664 | -111
java.lang.management.ThreadInfo | -1056 | -11
java.lang.Object[] | -19000 | -598
java.lang.reflect.Constructor | 4960 | 62
java.lang.reflect.Constructor[] | 1432 | 64
java.lang.reflect.Field | 30888 | 429
java.lang.reflect.Field[] | 2464 | 43
java.lang.reflect.Method | 33528 | 381
java.lang.reflect.Method[] | 6496 | 117
java.lang.String | 7824 | 326
java.lang.String[] | 5832 | 138
java.lang.StringBuilder | -1296 | -54
java.util.Collections$UnmodifiableMap | -634624 | -19832
java.util.HashMap | -909744 | -18953
java.util.HashMap$Node[] | 3328 | -13
java.util.LinkedHashMap$Entry | 23560 | 589
java.util.TreeMap | -1152 | -24
java.util.TreeMap$Entry | -9160 | -229
java.util.TreeMap$EntryIterator | -1024 | -32
java.util.TreeMap$KeyIterator | -1664 | -52
org.jboss.jandex.ClassInfo | 158656 | 0
org.jboss.jandex.FieldInternal[] | -98704 | -6169
org.jboss.jandex.MethodInternal[] | -9168 | -573

MikeEdgar:
PR branch re-based on master following release 2.2.3.

src/test/java/org/jboss/jandex/test/ClassInfoMemberPositionTestCase.java
```diff
@@ -0,0 +1,679 @@
...
+import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.Index;
+import org.jboss.jandex.IndexReader;
+import org.jboss.jandex.IndexWriter;
+import org.jboss.jandex.Indexer;
+import org.jboss.jandex.MethodInfo;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ClassInfoMemberPositionTestCase {
```
n1hility, line 42: Nice test!

n1hility (approved):

n1hility:
I am working on getting 3.0 out the door soon. This change looks good other than the one nit, which can be fixed in a subsequent change.

Issue #67, opened 2019-11-07
IndexView - add method/s to handle repeatable annotations


mkouba:
`org.jboss.jandex.IndexView.getAnnotations(DotName)` currently does not return the instances of repeatable annotations and that's ok (repeatable annotations are compiled into the container annotations anyway). However, we should probably add some convenient method to handle repeatable annotations as well, ie. something similar to `java.lang.reflect.AnnotatedElement.getAnnotationsByType(Class<T>)`. Of course, we can't modify the behavior of `IndexView.getAnnotations()` due to backwards compatibility.

Example:
```java
@Repeatable(Routes.class)
public @interface Route {

    @interface Routes {
        Route[] value();
    }
}

// the boolean param can be used to specify whether to return repeatable annotations or not  
Collection<AnnotationInstance> annotations = index.getAnnotations(DotName.createSimple(Route.class.getName()), true); 
```

FroMage:
+1 to handling this automatically. I think we should be able to just ignore the container annotation at index-time if we know it's a container annotation and make the API return the right thing for `index.getAnnotations(ROUTE_DOTNAME)` as well as `index.getAnnotatedClasses(ROUTE_DOTNAME)`.

If not, we need a new API for both use-cases (getting annotations from an annotated element, and getting the annotated element from an annotation name).

mkouba:
So while playing with a POC I've just realized that if we add `IndexView.getAnnotations(DotName, IndexView)` that would handle repeatable annotations we would have to copy the annotation instances and set the target from the container because nested ``AnnotationInstance``s return `null` from the `org.jboss.jandex.AnnotationInstance.target()` and thus would be unusable. @n1hility is that OK?

PR #79, opened 2020-04-06
Repeatable annotations

mkouba:
This is a POC for #67.

I've added several methods, most of them follow the naming and approach of `AnnotationInstance.valueWithDefault()`, i.e. the method declares an `IndexView` param that is used to obtain the annotation type definition and determine the repeatable annotation container (if present).

Note that `MethodInfo.annotationsWithRepeatable()` is a bit special because it should align with the contract of `MethodInfo.annotations()` which returns annotations declared on method or its parameters.

`IndexView.getAnnotationsWithRepeatable()` and `MethodInfo.annotationsWithRepeatable()`  are also special because they need to replace the target of the returned annotation instances for the ones coming from the container, otherwise it wouldn't be possible to determine the exact location of the respective annotation (see `org.jboss.jandex.AnnotationInstance.target()`).

NOTE: `annotationsWithRepeatable()` is not very nice but I did not find a better name that would indicate the meaning...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,4 +149,17 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <repositories>
+        <repository>
+            <id>jboss-public-repository</id>
+            <name>JBoss Repository</name>
+            <url>https://repository.jboss.org/nexus/content/groups/public</url>
+            <releases>
+                <enabled>true</enabled>
+            </releases>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+        </repository>
+    </repositories>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -131,7 +131,7 @@
         <dependency>
             <groupId>org.jboss.jandex</groupId>
             <artifactId>typeannotation-test</artifactId>
-            <version>1.0</version>
+            <version>1.1</version>
             <type>jar</type>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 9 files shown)


mkouba:
@n1hility @FroMage CC

mkouba:
Yeah, that was my original proposal - the names without the `WithRepeatable` suffix. But it's not very obvious. And `getAnnotationsRecursive()` is better but again it's not very explicit and I'd like to avoid the complexity of the javadoc from `java.lang.reflect.AnnotatedElement` (present, indirectly present, associated, etc.). Also I think that the original methods should not be deprecated because they're legal and easier to use for most use cases.

n1hility:
@mkouba this PR looks good to me

mkouba:
@n1hility Ah, do we really still target java 6?

mkouba:
Repeatable annotations are supported in jdk8+ so we wouldn't be able to write a proper test...

n1hility:
@mkouba I know we chatted about dropping everything before 8, but until that happens why don't we just add an example to here: https://github.com/wildfly/typeannotation-test it's already used for testing by jandex. This is how type annotation testing is done currently.

mkouba:
I'm not a big fan of externalizing the test suite unless really necessary. I'll add a special profile (and test folder) that will only activate if JDK8+ is used.

mkouba:
@n1hility the PR should be ready now...

n1hility:
> I'm not a big fan of externalizing the test suite unless really necessary. I'll add a special profile (and test folder) that will only activate if JDK8+ is used.

I think the way this is done it might force the jandex artifact and not just the test to be java 8. I will need to look into this.

mkouba:
@n1hility You can always disable the profile explicitly with `-P !jdk8`.

n1hility:
@mkouba I hope you don't mind but I went ahead and flipped this to use typeannotation-test, so that everything works consistently. The test suite is not actually external, the jdk8+ types are just example classes, and to jandex those are really just data.

mkouba:
No problem. Thanks!

PR #82, opened 2020-06-17
Override equals()/hashCode() in AnnotationValue.BooleanValue

### LINKED ISSUE #81: org.jboss.jandex.AnnotationValue.BooleanValue does not override equals()/hashCode()
As a result, the boolean values is ignored in `equals()`, i.e. it returns always `true` for two boolean annotation values with the same name.

mkouba:
- resolves #81

### DIFF
--- a/src/main/java/org/jboss/jandex/AnnotationValue.java
+++ b/src/main/java/org/jboss/jandex/AnnotationValue.java
@@ -1036,6 +1036,28 @@ public abstract class AnnotationValue {
             return value;
         }
 
+        @Override
+        public int hashCode() {
+            int result = super.hashCode();
+            result = 31 * result + Boolean.hashCode(value);
+            return result;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj) {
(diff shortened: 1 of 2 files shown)


mkouba:
CC @n1hility 

Maybe it's a good time to release another bugfix version of jandex too?

famod:
Have you seen https://github.com/wildfly/jandex/issues/81#issuecomment-645210224 ?

mkouba:
@famod Yes, I did. I think that we should keep things simple and use the "specialized" versions of `equals()/hashCode()`. This way we can avoid unnecessary boxing for primitive values and apply specific strategies for `hashCode()`.

famod:
@mkouba Ok, your decision (no offense!).
It's just that the abstract base class provides a `toString()` that is inherently incomplete and AFAICS _no_ subclass should rely on that (but is not foced to implement a complete `toString()`).

mkouba:
> Ok, your decision (no offense!).

It's up to @n1hility. I've only sent a PR ;-).

Issue #84, opened 2020-08-13
Index classes used in the constant pool


FroMage:
Issue #50 was flawed because we didn't want to index all methods calls, so I thought that clients could specify which calls we wanted to flag during indexing, but this doesn't work with packaged indexes.

Now, in Quarkus we have a double step to run transformers conditionally:

- We have an APT plugin in some modules that add a marker file to `META-INF`
- If the marker is here, we register transformers for all class files of that jar, but only if their constant pool references class constants from a list `Foo,Bar,Gee`

This allows us to only transform classes that use a certain special feature.

Now, in https://github.com/quarkusio/quarkus/issues/814 I want to add a transformer based on the usage of a `Router` class, and could take the same road, but it would still require scanning the constant pool of every class. I would either create a new extension just for this, or add it to the existing resteasy extension, but every user would pay the ConstPool scanning.

And in https://github.com/quarkusio/quarkus/issues/10929 I definitely don't want to create an extension for a single class, which should be provided by Quarkus Core. And I don't want to add it to Quarkus Core if it means scanning every application class ConstPool even if nobody uses this class. I can't use the marker file in this case.

Therefore, I think we can solve all issues if Jandex indexes `CONSTANT_Class (7)` entries. Ultimately all method calls reference Fieldref/Methodref/InterfaceMethodref (https://docs.oracle.com/javase/specs/jvms/se14/html/jvms-4.html#jvms-4.4.2) which reference CONSTANT_Class_info (https://docs.oracle.com/javase/specs/jvms/se14/html/jvms-4.html#jvms-4.4.1).

I don't think this would add a huge size to the index, and would allow us to scan efficiently for classes which make calls to special classes, which would work for ORM with Panache, as well as the Reverse-Router and Logging with Panache use-cases without having to use marker files or even creating extensions for single classes just to have the marker.

I can work on this if you think this is the right idea.

WDYT @n1hility @stuartwdouglas @mkouba ?

stuartwdouglas:
I think the idea is ok. My gut feeling is that it won't bloat the index size too much, as we should already have DotName objects for a lot of the classes.

n1hility:
Yeah lets look at some numbers on this. I think its a good idea.

n1hility:
@FroMage do you have the bandwidth to do a small prototype?

FroMage:
I have an implementation, I tried it on indexing Jandex itself, and got these index sizes:

```
Indexed 145 classes
V9 size: 774 bytes
[...]

PR #85, opened 2020-08-19
WIP Added class constant indexing for #84

FroMage:
So I index all types used in the Constant Pool as Class Constants, and make a Map from the used type to the user of the type as a `ClassInfo`.

I thought about mapping to `DotName` but given that they can only belong to the index (since we indexed them), it's not clear that it's worth it.

I thought about mapping to a `Set<ClassInfo>` but that type doesn't implement hashCode/equals and so it was lying. In the end I've gone like most parts of the API (I did see a `Set<ClassInfo>` somewhere but it was alone) and used `List` even though there are no duplicates. At least the type doesn't lie about unicity, though it lies about ordering.

I've a question wrt `Index.create()` which is a `static public` method, so probably I'll need an overload for backwards compat, to accomodate for the new extra param, right?

As for the index format, I've bumped to V10, and tucked a `Map<DotName,List<DotName>>` right after the type list.

This should be all tested.

I'll add docs and remove the index size test once we agree on how to proceed with this PR.

Let me know what you think.

### DIFF
--- a/src/main/java/org/jboss/jandex/CompositeIndex.java
+++ b/src/main/java/org/jboss/jandex/CompositeIndex.java
@@ -219,5 +219,25 @@ public class CompositeIndex implements IndexView {
         }
         return Collections.unmodifiableCollection(allKnown);
     }
+    
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Collection<ClassInfo> getKnownUsers(DotName usedClass) {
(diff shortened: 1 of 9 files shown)


FroMage:
Damnit, it should have been a draft PR but I can't seem to be able to change this after the fact…

FroMage:
Ah, found it :)

n1hility:
Superseded  by #112

PR #104, opened 2021-01-16
Add GitHub Action to perform regression testing with Apache Calcite bytecode

vlsi:


vlsi:
@n1hility , I think the PR is ready. Well, regression jobs fail, however, they highlight the issues.

.github/workflows/ci.yml
```diff
@@ -20,6 +20,53 @@ jobs:
...
+    strategy:
+      matrix:
+        # javac 8 and 11 might produce slightly different bytecode
+        java-version: [8, 11]
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 50
+    - name: 'Set up JDK 11'
+      uses: actions/setup-java@v1
```
vlsi, line 39: Just in case, if you prefer `joschi/setup-jdk`, then this should probably be replaced.

n1hility:
@vlsi I fixed a few more issues that were reported in the logs here, but I was having trouble reproducing the failure locally (downloading the classes that were built and then running jandex on them had no issues). Since this error looks like one of the ones that was fixed, I thought I saw a caching issue but I verified at least part of the build is picking up changes. Not sure if there is anything special about the usage of your gradle plugin but it didn't look like it glancing at the source. Anyway I need to crash but let me know if you spot anything.

vlsi:
Oh, let me see.

vlsi:
@n1hility , 
1) Gradle does not cache the results of failure tasks (it is not implemented in the current Gradle versions)
2) I can reproduce `Required class information is missing` failures locally.

For instance, I download `calcite-work.tgz`, I take `calcite/core/build/classes/java/main/org/apache/calcite/adapter/clone/ArrayTable$1.class` file (1477 bytes), and I run the following:

```
public class TestClassBounds {
    @Test
    public void classBounds() throws IOException {
        ClassInfo info = indexer.index(new FileInputStream("/tmp/ArrayTable$1.class"));
        System.out.println("info = " + info);
        System.out.println("subclasses");
        index.printSubclasses();
        System.out.println("annotations");
        index.printAnnotations();
    }
}
```

The error can be reproduced as follows: https://github.com/wildfly/jandex/issues/92

Are you sure you already fixed the error?

256fae49ac05d6303b5ba7b75a75a5b2169c748b yields 

```
java.lang.IllegalStateException: Required class information is missing

	at org.jboss.jandex.Indexer.rebuildNestedType(Indexer.java:989)
	at org.jboss.jandex.Indexer.resolveTypePath(Indexer.java:842)
	at org.jboss.jandex.Indexer.resolveTypeAnnotation(Indexer.java:718)
	at org.jboss.jandex.Indexer.resolveTypeAnnotations(Indexer.java:619)
	at org.jboss.jandex.Indexer.index(Indexer.java:1678)
	at org.jboss.jandex.test.TestClassBounds.classBounds(TestClassBounds.java:15)
```

n1hility:
@vlsi thanks for looking and the pointer. For some reason I still can't reproduce on my local setup (mac and linux) [both using the test and the exact class you downloaded] so this is an odd one. I'll dig further!

n1hility:
Wow it even passes running as GitHub action through the jandex maven run using your exact code and class:
https://github.com/n1hility/jandex/runs/1744092156#step:4:1087

I'll try and kick of the gradle build locally and see if I can get it to fail and debug.

n1hility:
I finally have a local scenario, looking into it.

vlsi:
Does that mean the error reproduces in ceratin JVMs only?

n1hility:
@vlsi yes, so far it looks like it is a compiler bug with anonymous classes that affects some range (don't know yet) of 8 versions. it's a bit of a comedy of errors. In a few cases I had also copied the file over without properly escaping $ So ArrayTable$1.class ended up  matching ArrayTable.class, yet when i was disassembling with javap I had escaped it and didn't notice the difference. In other cases I had built with java 11 and it worked.

n1hility:
oh yeah and then i lost my internet connection. so this bug is resisting being fixed :)

n1hility:
@vlsi We now have a clean run! 🎉

n1hility:
@vlsi I have released 2.2.3.Final including all of these fixes.

vlsi:
@n1hility , thanks for the fast release. I'll reiterate with Calcite as the release appears on OSSRH (it is not there yet: https://repo.maven.apache.org/maven2/org/jboss/jandex/ ).

I wonder if you have the changelog somewhere.
It might be helpful to put some notes there, especially taking into account the nature of the issues was moot and the exceptions were quite puzzling.

I guess we'll need to convince @vladmihalcea to bump Hibernate's `jandex` dependency to prevent the users from running into issues like https://stackoverflow.com/a/65853462/1261287.

vladmihalcea:
@vlsi I'm no longer working on the Hibernate project, so you are better off talking to @Sanne about Hibernate issues.

Sanne:
Good idea, thanks!
 - https://hibernate.atlassian.net/browse/HHH-14415

PR #105, opened 2021-01-16
Fix IllegalArgumentException: Not an array type for method like void method(@Nullable Object[])

vlsi:
Fixes https://github.com/wildfly/jandex/issues/102

Note: there are lots of ways to structure the project, so I would be glad to hear suggestions.
Currently `jandex` is top-level `packaging=bundle`, which limits Maven's ability to have `modules`.

However, even the structure I suggest enables me to open both projects in IDEA, run test, debug project, etc.


n1hility:
@vlsi Thanks for the contribution! I did a small refactor to also cover inner class params: https://github.com/n1hility/jandex/commit/c3d0cc87c81b83915d50429b9d0537ccca0979bc

I temporarily removed the GitHub action and replaced the java8 test module vs the current approach since these are currently nonfictional and the intention is to fix the bug in a current release quickly. 

I didn't update your PR quite yet since I wanted to give you a chance to point out any concerns/issues. The other generics issue I will post up tomorrow, your fix is correct but I just need to verify the compatibility impact.

n1hility:
Ok this PR has been updated with the commits I linked. If you have any concerns let me know, otherwise I plan to merge these tomorrow.

src/main/java/org/jboss/jandex/Indexer.java
```diff
@@ -747,6 +757,30 @@ private void resolveTypeAnnotation(AnnotationTarget target, TypeAnnotationState
...
 
+    private boolean skipBridge(TypeAnnotationState typeAnnotationState, MethodInfo method) {
+        // javac copies annotations to bridge methods (which is good), however type annotations
+        // might become invalid. For instance, the bridge signature for @Nullable Object[] is
+        // Object (non-array), so usage=ARRAY signature is invalid. Other cases include annotated
+        // inner classes.
+        //
+        // We ignore those annotations for the bridge methods.
+        // See https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6695379
+        return isBridge(method) && typeAnnotationState.bridgeIncompatible;
```
vlsi, line 768:
```suggestion
        return typeAnnotationState.bridgeIncompatible && isBridge(method);
```

The reorder of the conditions might slightly improve the performance  since `bridgeIncompatible` is slightly easier to verify

vlsi:
> removed the GitHub action

This is sad since currently, it is hard to see if the fix works for Calcite or not.

Issue #80, opened 2020-04-20
ArrayIndexOutOfBoundsException


diamondq:
Ran into a case where scanning some libraries found an ArrayIndexOutOfBoundsException. This was based on 2.1.3.Final.

```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 1
	at org.jboss.jandex.Indexer.updateTypeTarget(Indexer.java:841)
	at org.jboss.jandex.Indexer.updateTypeTargets(Indexer.java:624)
	at org.jboss.jandex.Indexer.index(Indexer.java:1603)
	at test.Main.main(Main.java:14)
```
This occurred using a simple test program:
```
package test;

import java.io.IOException;
import java.io.InputStream;

import org.jboss.jandex.Index;
import org.jboss.jandex.Indexer;

public class Main {

  public static void main(String[] args) throws IOException {
    Indexer indexer = new Indexer();
    InputStream stream = Main.class.getClassLoader().getResourceAsStream("test/Broken.class");
    indexer.index(stream);
    Index index = indexer.complete();
  }
}
```
and failed on this class:
```
package test;

import java.util.List;

import org.checkerframework.checker.nullness.qual.NonNull;

public class Broken {

  public static <U extends @NonNull List<String>> U listOf() {
    return null;
  }

}
```
This was, obviously, from a much more complicated example, but I tried to reduce it down to the simplest case that still generated the exception.

As far as I could tell, the problem lies here:
```
            case TYPE_PARAMETER_BOUND: {
                TypeParameterBoundTypeTarget boundTarget = target.asTypeParameterBound();
                type = getTypeParameters(enclosingTarget)[boundTarget.position()]
                           .asTypeVariable().boundArray()[boundTarget.boundPosition()];
                break;
            }
```
where the boundTarget.position is 1, and boundPosition is 1, but while the getTypeParameters returns a 2-element array, the second array (for boundPosition) is only a 1-element array, and thus, fails.
Unfortunately, I couldn't follow the code logic enough to track the underlying problem, so just reporting with all the data I did get.

Cheers.

vlsi:
>Unfortunately, I couldn't follow the code logic enough to track the underlying problem, so just reporting with all the data I did get.

I explained the issue in https://github.com/wildfly/jandex/issues/99
The key point is that `U extends @NonNull List<String>` compiles to `U::List...` signature, so the bound indexing starts with 1, and the position of 0 is omitted.

[...]

PR #106, opened 2021-01-16
Handle optional class bound in the class type variables

### LINKED ISSUE #99: Incorrect handling of the optional class bound in the class type variables
```java
package org.jboss.jandex.test;

import java.lang.annotation.ElementType;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;
import java.lang.annotation.Target;
import java.math.BigInteger;

public class ClassBounds<V1 extends @ClassBounds.NonNull BigInteger,
        V2 extends @ClassBounds.Nullable Runnable> {
    @Retention(RetentionPolicy.RUNTIME)
    @Target(ElementType.TYPE_USE)
    public @interface NonNull {
    }

    @Retention(RetentionPolicy.RUNTIME)
    @Target(ElementType.TYPE_USE)
    public @interface Nullable {
    }
}
```

Test class:

```java
package org.jboss.jandex.test;

import org.jboss.jandex.Index;
import org.jboss.jandex.Indexer;
import org.junit.Test;

import java.io.IOException;

public class TestClassBounds {
    @Test
    public void classBounds() throws IOException {
        indexer.index(ClassBounds.class.getResourceAsStream("ClassBounds.class"));
        System.out.println("subclasses");
        index.printSubclasses();
        System.out.println("annotations");
        index.printAnnotations();
    }
}
```

Actual error:

```
java.lang.ArrayIndexOutOfBoundsException: 1
	at org.jboss.jandex.Indexer.updateTypeTarget(Indexer.java:847)
	at org.jboss.jandex.Indexer.updateTypeTargets(Indexer.java:630)
	at org.jboss.jandex.Indexer.index(Indexer.java:1618)
	at org.jboss.jandex.test.TestClassBounds.classBounds(TestClassBounds.java:13)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
```

---

It looks like jandex misses something like https://github.com/openjdk/jdk/commit/9d7f8bc9a053595c290ae339b902dc762b848ab9

vlsi:
See https://github.com/openjdk/jdk/commit/9d7f8bc9a053595c290ae339b902dc762b848ab9
See https://github.com/wildfly/jandex/issues/99

fixes #99, #80


src/main/java/org/jboss/jandex/GenericSignatureParser.java
```diff
@@ -406,6 +406,11 @@ private Type parseTypeParameter() {
         // Class bound has an optional reference type
         if (signature.charAt(pos) != ':') {
             bounds.add(parseReferenceType());
+        } else {
+            // Pretend optional reference type is Object
+            // See https://github.com/openjdk/jdk/commit/9d7f8bc9a053595c290ae339b902dc762b848ab9
+            // See https://github.com/wildfly/jandex/issues/99
+            bounds.add(ClassType.OBJECT_TYPE);
```
vlsi, line 413: By the way, I don't know if this has implications, however, it seems to be the least evil. Otherwise, jandex index format must be improved to support the distinction between `<T extends Object & Interface>` vs `<T extends Interface>` (which have `T:Object:Interface` and `T::Interface` signatures).

However, I guess inserting `Object` here allows accessing the annotations without updating the index format (e.g. to account for optional yet absent class bound).

My best guess is that optional class bounds are allowed here to allow users to configure the erasure signatures.

For instance `<T extends Number>` erases to `Number` while `<T extends Object & Number>` erases to `Object` which allows Java programmres to evolve old code (e.g. keep backward-compatible API while adding generics to their code).

^^^ The above is a wild guess, so I am not sure if it should be added to the code comment.

n1hility: So I agree it's a good solution: there is something to be said for having consistent  semantic representation vs the multiple encodings. However, since I would like to include this fix as a bug fix release, I was a little worried about behavioral compatibility differences that might impact some users. For example, previous code might have expected T extends Runnable to have .bounds().get(0) return Runnable, whereas after this change it would be Object.  So I changed the approach to instead use a hidden marker flag only for the indexing process, and in that way the implied object class bound is hidden. 

Your wild guess is correct, the JLS does define erasure to be the left most class bound, so Object & Interface is functionally different than just Interface.

I have updated the PR with my changes. I intend to merge tomorrow provided you have no concerns or questions with the update

src/main/java/org/jboss/jandex/TypeVariable.java
```diff
@@ -131,19 +142,19 @@ TypeVariable copyType(int boundIndex, Type bound) {
...
+        int hash = this.hash & Integer.MAX_VALUE;
         if (hash != 0) {
             return hash;
         }
 
         hash = super.hashCode();
         hash = 31 * hash + name.hashCode();
         hash = 31 * hash + Arrays.hashCode(bounds);
-        return this.hash = hash;
+        return this.hash |= (hash & Integer.MAX_VALUE);
```
vlsi, line 158: What is the purpose of the change?

src/main/java/org/jboss/jandex/TypeVariable.java
```diff
@@ -131,19 +142,19 @@ TypeVariable copyType(int boundIndex, Type bound) {
...
         bounds[boundIndex] = bound;
-        return new TypeVariable(name, bounds, annotationArray());
+        return new TypeVariable(name, bounds, annotationArray(), hasImplicitObjectBound());
     }
 
     @Override
     public int hashCode() {
-        int hash = this.hash;
+        int hash = this.hash & Integer.MAX_VALUE;
         if (hash != 0) {
```
vlsi, line 151: Frankly speaking, this `MAX_VALUE` vs `MIN_VALUE` looks very hard to review.

What do you think:

a) If a regular `boolean` field was added? It would be way easier to understand.
b) Another option is to introduce a named constant for the bit. Currently, MIN vs MAX says nothing with regards to `implicitObjectBound`

n1hility: Ideally we would be able to use a boolean but unfortunately it costs us another 8 bytes per type variable, and these can easily be in 10-100k range of instances in large projects. Openjdk on x86-64 aligns objects on 8 byte boundary and sometimes you can get lucky if there is unused padding. However I checked and adding a boolean moves us from 32 bytes to 40 bytes per TypeVar (shallow space). Good point on b) I introduced intermediary constants

PR #112, opened 2021-01-25
Add class constant indexing

### LINKED ISSUE #84: Index classes used in the constant pool

Ladicek:
- resolves #84

### DIFF
--- a/src/main/java/org/jboss/jandex/Main.java
+++ b/src/main/java/org/jboss/jandex/Main.java
@@ -87,7 +87,7 @@ public class Main {
         Result result = (source.isDirectory()) ? indexDirectory(source, indexer) : JarIndexer.createJarIndex(source, indexer, outputFile, modify, jarFile, verbose);
 
         double time = (System.currentTimeMillis() - start) / 1000.00;
-        System.out.printf("Wrote %s in %.4f seconds (%d classes, %d annotations, %d instances, %d bytes)\n", result.getName(), time, result.getClasses(), result.getAnnotations(), result.getInstances(), result.getBytes());
+        System.out.printf("Wrote %s in %.4f seconds (%d classes, %d annotations, %d instances, %d class usages, %d bytes)\n", result.getName(), time, result.getClasses(), result.getAnnotations(), result.getInstances(), result.getUsages(), result.getBytes());
         return result.getIndex();
     }
 
--- a/src/main/java/org/jboss/jandex/CompositeIndex.java
+++ b/src/main/java/org/jboss/jandex/CompositeIndex.java
@@ -219,5 +219,25 @@ public class CompositeIndex implements IndexView {
         }
         return Collections.unmodifiableCollection(allKnown);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Collection<ClassInfo> getKnownUsers(DotName className) {
+        final List<ClassInfo> users = new ArrayList<ClassInfo>();
+        Set<DotName> processedClasses = new HashSet<DotName>();
+        for (IndexView index : indexes) {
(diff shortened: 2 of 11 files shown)


Ladicek:
This supersedes #85.

It also contains an extra commit that attempts to improve index writer performance, which is in direct conflict with #107. If you prefer #107, just let me know and I'll drop the commit from here.

src/main/java/org/jboss/jandex/Main.java
```diff
@@ -87,7 +87,7 @@ private Index getIndex(long start) throws IOException {
         Result result = (source.isDirectory()) ? indexDirectory(source, indexer) : JarIndexer.createJarIndex(source, indexer, outputFile, modify, jarFile, verbose);
 
         double time = (System.currentTimeMillis() - start) / 1000.00;
-        System.out.printf("Wrote %s in %.4f seconds (%d classes, %d annotations, %d instances, %d bytes)\n", result.getName(), time, result.getClasses(), result.getAnnotations(), result.getInstances(), result.getBytes());
+        System.out.printf("Wrote %s in %.4f seconds (%d classes, %d annotations, %d instances, %d class usages, %d bytes)\n", result.getName(), time, result.getClasses(), result.getAnnotations(), result.getInstances(), result.getUsages(), result.getBytes());
```
aimiebell, line 90: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

Ladicek:
Hi @n1hility, any plans around Jandex 3? :-)

PR #114, opened 2021-03-16
bitsToLong given wrong result when the bytes having the highest bit set

qiuhw:
The bitsToLong method of Indexer class given wrong result when the bytes having the highest bit set. For example when I have a parameter annotation `@Max(19999999999L)`, the method converted it to `-1474836481L`. This pull request fix that problem.


n1hility:
Oops yes the <<24 will sign extend when widened

PR #118, opened 2021-05-06
Replace original positions binary search with `IdentityHashMap`

MikeEdgar:
@n1hility - thanks for pointing me in the right direction in your comment on #73. I took a look and it turns out that the binarySearch approach is in fact broken as it relates to bridge methods. I ran into some differences in the index testing with an `IdentityHashMap` and I believe it is a necessary change.

Consider class `com.sun.xml.internal.bind.v2.model.impl.RuntimeBuiltinLeafInfoImpl.UUIDImpl` with the following methods:
```
void                   <init>()
java.util.UUID         parse(java.lang.CharSequence text) throws org.xml.sax.SAXException
java.lang.Object       parse(java.lang.CharSequence)      throws com.sun.xml.internal.bind.api.AccessorException, org.xml.sax.SAXException
java.lang.String       print(java.lang.Object)            throws com.sun.xml.internal.bind.api.AccessorException
java.lang.CharSequence print(java.lang.Object)            throws com.sun.xml.internal.bind.api.AccessorException
java.lang.String       print(java.util.UUID v)]
```

Two of the `print` methods are identical in name and parameter type/name, resulting in incorrect results being placed in the original positions array. I did not notice any significant runtime performance difference between this PR and the original version when indexing JRE 8's rt.jar.

### DIFF
--- a/src/main/java/org/jboss/jandex/ClassInfo.java
+++ b/src/main/java/org/jboss/jandex/ClassInfo.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


n1hility:
@MikeEdgar Thanks! The issue you uncovered is interesting. I might need to look into the comparison contract.

PR #119, opened 2021-05-06
Support modules and records

MikeEdgar:
Adds support for the scanning of `module-info` class files (JDK 9+) as well as annotations on record components (GA JDK 16).

- New `ModuleInfo` model referenced from its parent `ClassInfo`'s `nestingInfo`. Annotations from the Java `module` are still present on the `ClassInfo`, but having the `ModuleInfo` model with the descriptor details is a "nice to have" to support other potential use cases
- Index modules separately from classes
- Add `AnnotationTarget.Kind.RECORD_COMPONENT`
- Add new `RecordComponentInfo` implementation of `AnnotationTarget`. Annotations on record components can be accessed very similarly to the way annotations on method parameters are accessed. I.e., retrieve the class's annotations and identify those with a `target.kind()` of `Kind.RECORD_COMPONENT`. The component's associated `ClassInfo` and `FieldInfo` can be obtained directly from `RecordComponentInfo`

If this approach looks reasonable/acceptable, I can also go back and add JavaDocs where missing and do some additional research/optimizations if necessary. Note that the build will fail prior to a release of typeannotation-test with the changes from wildfly/typeannotation-test#1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,6 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>2.1</version>
                 <configuration>
                     <debug>true</debug>
                     <source>1.6</source>
@@ -89,7 +88,6 @@
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
-                <version>2.3.7</version>
                 <extensions>true</extensions>
                 <configuration>
                     <archive>
--- a/src/main/java/org/jboss/jandex/AnnotationTarget.java
+++ b/src/main/java/org/jboss/jandex/AnnotationTarget.java
@@ -58,7 +58,14 @@ public interface AnnotationTarget {
         /**
          * An object of type {@link org.jboss.jandex.TypeTarget}
          */
-        TYPE}
+        TYPE,
+
+        /**
+         * An object of type {@link org.jboss.jandex.RecordComponentInfo}
+         * @since 2.4
+         */
+        RECORD_COMPONENT
+    }
 
     /**
      * Returns the kind of object this target represents.
@@ -107,5 +114,13 @@ public interface AnnotationTarget {
       * @since 2.0
       */
     TypeTarget asType();
+
+    /**
+     * Casts and returns this target as a <code>RecordComponentInfo</code> if it is of kind <code>RECORD_COMPONENT</code>
+     *
+     * @return this instance cast to a record component
+     * @since 2.4
+     */
+    RecordComponentInfo asRecordComponent();
 
 }
--- a/src/main/java/org/jboss/jandex/ClassInfo.java
+++ b/src/main/java/org/jboss/jandex/ClassInfo.java
@@ -50,6 +50,7 @@ import java.util.Map;
  */
 public final class ClassInfo implements AnnotationTarget {
 
+    private static final int MODULE = 0x8000;
     private static final int MAX_POSITIONS = 256;
     private static final byte[] EMPTY_POSITIONS = new byte[0];
 
@@ -63,8 +64,10 @@ public final class ClassInfo implements AnnotationTarget {
     private Type[] typeParameters;
     private MethodInternal[] methods;
     private FieldInternal[] fields;
+    private RecordComponentInternal[] recordComponents;
     private byte[] methodPositions = EMPTY_POSITIONS;
     private byte[] fieldPositions = EMPTY_POSITIONS;
+    private byte[] recordComponentPositions = EMPTY_POSITIONS;
     private boolean hasNoArgsConstructor;
     private NestingInfo nestingInfo;
 
@@ -80,12 +83,19 @@ public final class ClassInfo implements AnnotationTarget {
         LOCAL,
 
         /** An unnamed class enclosed within a code block */
-        ANONYMOUS}
+        ANONYMOUS
+    }
 
     private static final class NestingInfo {
         private DotName enclosingClass;
         private String simpleName;
         private EnclosingMethodInfo enclosingMethod;
+
+        // non-null if the class is in fact a module descriptor
+        // this field would naturally belong to ClassInfo, but is present here
+        // to make the ClassInfo object smaller in the most common case:
+        // non-nested class that is not a module descriptor
+        private ModuleInfo module;
     }
 
     /**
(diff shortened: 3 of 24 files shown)


n1hility:
hey @MikeEdgar I would like to switch the typeannotation-test module to the Apache License. Do you agree to placing your recent contributions to that project under those terms?

MikeEdgar:
> hey @MikeEdgar I would like to switch the typeannotation-test module to the Apache License. Do you agree to placing your recent contributions to that project under those terms?

@n1hility - yes I agree those terms.

n1hility:
@MikeEdgar awesome thanks!

MikeEdgar:
Checking on the NPEs...

MikeEdgar:
It's not getting the correct `module-info` from the class path on JDK 11 for the unit test. Looking at a fix...

MikeEdgar:
I updated the test to ensure only the module-info it's expecting will be indexed. I'm happy to do it some other way if you can think of something more elegant. Apparently I only ran the tests on JDK 8 :-(

n1hility:
> I updated the test to ensure only the module-info it's expecting will be indexed. I'm happy to do it some other way if you can think of something more elegant. Apparently I only ran the tests on JDK 8 :-(

Out of curiosity did the NPE indicate an internal bug with processing a particular JDK module? Im just wondering if this uncovered an unrelated issue.

MikeEdgar:
Yes, I think so. The module version may not be consistently set. I'll add in the null checks to the writer to account for that.

MikeEdgar:
The latest commit adds some null checking and also indexes all modules found on the test class's class path. That gives some additional variety including some with `null` versions and `requires` versions.

MikeEdgar:
Rebased and squashed the PR. Let me know if you think anything else looks like an issue.

Ladicek (changes requested):
I personally think this shouldn't go in as is. The modules support seems fine, but the records support is very limited (e.g. there's no `ClassInfo.recordComponent[s]`). I took the liberty of refactoring that significantly, to bring records support in line with fields and methods. Here's my branch, on top of this PR: https://github.com/Ladicek/jandex/commits/modules-records-refactoring

MikeEdgar:
> the records support is very limited (e.g. there's no `ClassInfo.recordComponent[s]`)

I can't say that I disagree :-) I implemented it that way (i.e. similar to method parameter annotations) to stay sensitive to memory utilization, but the API additions you propose are certainly better.

Ladicek:
I just added this to the 2.4.0.Final milestone (we recently agreed that what was originally referred to as 3.0 can and should be released as 2.4, because there's nothing breaking and we need to do some breaking changes later). I think we agreed with @n1hility that this is the last thing to get into 2.4.

So we need to finalize this somehow :-) @MikeEdgar WDYT? Should I add a commit to your branch, or amend the commit on your branch, or submit a new PR with a single commit? (I don't think this should go in as 2 commits, at least not as the 2 commits I have in my branch: https://github.com/Ladicek/jandex/commits/modules-records-refactoring. If we had 1 commit for modules and 1 commit for records, that would be fine, but I don't feel like separating the 2 things out of the existing 1 commit now :-) )

MikeEdgar:
@Ladicek - feel free to amend the commit in this PR. That might be the easiest and you should have access if you're a maintainer.

Ladicek:
I'll do that, thank you!

Ladicek:
Before I force-push, I tried to do some measurements with `~/.m2/repository/org/hibernate/hibernate-core/5.5.4.Final/hibernate-core-5.5.4.Final.jar`. It has 4919 classes, no records and no modules (AFAIK :-) ). So I'm interested in the overhead of this feature -- which I assume mostly comes from my additions, where `ClassInfo` has a list of record components (in addition to fields and methods).

Index created with current Jandex master (d4f7f7c0dff7699c4e48fc938ca2b31d7eea567c) has 1_398_783 bytes, and when loaded to memory, its retained size is 6_216_504 bytes (according to Eclipse Memory Analyzer).

Index created with my latest changes (d307d03f8db01e5e834d7ea3dc00270e9fe368cc) has 1_418_501 bytes, and when loaded to memory, its retained size is 6_295_248 bytes (according to Eclipse Memory Analyzer).

That is, the difference on disk is 19_718 bytes, while the difference in memory is 78_744 bytes.

The index on disk newly contains, for each class, the number of record components (always 0, so stored as 1 byte, and no actual info about record components), and an index of the positions array in the byte array pool (it's an empty array, so let's assume the index is also serialized as 1 byte). For 4919 classes, that is 9_838 bytes, almost half of the difference.

The index in memory newly contains a bunch of references to empty arrays, but MAT shows that they all point to the same empty array instance. So I guess the difference is dominated by the object layout differences. `ClassInfo` gained 2 fields for records and 1 field for modules, which means 12 bytes per instance (assuming compressed oops). The index has 4919 `ClassInfo` objects, which means 59_028 bytes -- more than 3/4 of the difference.

There's no significant difference in indexing time, nor in index loading time.

Ladicek:
The memory measurement made me think: originally, the `ModuleInfo` reference was present in `NestingInfo`, and I moved it to `ClassInfo` because it has nothing to do with nesting. But the most common case is a top-level class that is not a module descriptor -- in which case, we waste 4 bytes (1 compressed reference) for nothing.

I tried to move the `ModuleInfo` reference back to `NestingInfo` and the retained size of the index decreased to 6_266_800 bytes (that is, 28_488 bytes out of the original 78_744 difference). Pretty nice! I'm adding that back, even if it's counter-intuitive. At least I'll add a comment.

MikeEdgar:
> the `ModuleInfo` reference was present in `NestingInfo`, and I moved it to `ClassInfo` because it has nothing to do with nesting

I was thinking about somehow making the `nestingInfo` field a generic Object reference that could either be a `NestingInfo` or a `ModuleInfo` since only one or the other should ever be used.

Ladicek:
Oh if only we had union types :-)

Ladicek:
@MikeEdgar @n1hility I'll let you review if you want (no pressure :rofl:), and then I think we're ready for 2.4.

MikeEdgar:
It may be worth mentioning that record components are also recorded as fields. Finding all of the annotations on the original record component will require looking at the annotations on both the field and the record component. Would it be worthwhile to have an API that somehow accounts for that, or even adds the annotations specific to the record components into the field's annotation array (which would remove the need to have a separate model for the record components)?

I avoided that originally to follow the model for method parameter annotations, but it seems more reasonable now, unless I'm overlooking something.

Ladicek:
My understanding is that if the annotation is `@Target(ElementType.RECORD_COMPONENT)`, the annotation will only be on the record component. If it's `@Target(ElementType.FIELD)`, it will only be on the field. If it's `@Target({ElementType.FIELD, ElementType.RECORD_COMPONENT})`, it will be on both. Meaning that the set of annotations on the record component and on the component field may be wildly different. I would prefer to not merge them for that reason.

MikeEdgar:
That is my understanding as well. I am thinking too specific to how I'd like to use them :-), so you're right about keeping them separate.

n1hility:
IIRC record components and fields are mutually exclusive per JLS. If so, It’s interesting they didn’t go down the path of modeling them as an extended field (just an extra accessor name attribute or something). Have you guys looked at the generated bytecode does it double declare the fields and components? If you call getFields in Java reflection on a record type do you get anything?

MikeEdgar:
I did a reflection test and getDeclaredField(s) does return the components. They are double declared in the bytecode also.

Ladicek (approved):

PR #121, opened 2021-05-27
Include relevant type-use annotations in MethodInfo/FieldInfo and fix index writing for TypeTarget

yrodiere:
See the reproducer for the detail of the problem I'm trying to fix. The problem was detected downstream in Quarkus: https://github.com/quarkusio/quarkus/issues/17491

See the message of the second-to-last commit for an explanation of the fix. Copying it here for convenience:

```quote
    Include relevant type-use annotations in MethodInfo/FieldInfo

    First, because that's what the javadoc of MethodInfo advertises.
    For example in the javadoc of
    org.jboss.jandex.MethodInfo#annotations(org.jboss.jandex.DotName):

         * Retrieves annotations declared on this method, by the name of the
           annotation. This includes annotations which are defined against
    method parameters, as
         * well as type annotations declared on any usage within the method
           signature. The <code>target()</code> of the returned annotation
    instances may be used to
         * determine the exact location of the respective annotation
           instance.

    Notice in particular "as well as type annotations declared on any usage
    within the method signature".

    The second, perhaps more important reason, is related to index
    writing/reading.

    In short, putting a reference to the annotation
    instance in the method means the annnotation instance will be
    written/read together with that method, with the (good) consequence
    that org.jboss.jandex.IndexWriterV2#writeTypeTarget will get passed the
    enclosing method as the "caller" parameter, and will now correctly set
    that method as the enclosing target.

    When the annotation instance was not referenced in the method, the
    annotation instance was written together with the enclosing type, so
    org.jboss.jandex.IndexWriterV2#writeTypeTarget got passed that type as
    the "caller" parameter and considered that the type was the enclosing
    target, which is wrong.
```

Note I had to change the project configuration to compile tests with a source/target set to `1.8`. Otherwise I just cannot add a reproducer, since type-use annotations were introduced in Java 8.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,7 +131,7 @@
         <dependency>
             <groupId>org.jboss.jandex</groupId>
             <artifactId>typeannotation-test</artifactId>
-            <version>1.6</version>
+            <version>1.8</version>
             <type>jar</type>
             <scope>test</scope>
         </dependency>
--- a/src/main/java/org/jboss/jandex/IndexReaderV2.java
+++ b/src/main/java/org/jboss/jandex/IndexReaderV2.java
@@ -644,6 +644,9 @@ final class IndexReaderV2 extends IndexReaderImpl {
         // This update is possible since annotations on a non-null target are unique and not shared
         for (AnnotationInstance annotation : annotations) {
             AnnotationTarget target = annotation.target();
+            if (target instanceof TypeTarget) {
+                target = ((TypeTarget) target).enclosingTarget();
+            }
             if (target instanceof MethodInfo) {
                 ((MethodInfo)target).setClassInfo(clazz);
             } else if (target instanceof MethodParameterInfo) {
--- a/src/main/java/org/jboss/jandex/Indexer.java
+++ b/src/main/java/org/jboss/jandex/Indexer.java
@@ -1258,7 +1258,9 @@ public final class Indexer {
             recordAnnotation(classAnnotations, annotationName, instance);
             recordAnnotation(masterAnnotations, annotationName, instance);
 
-            if (target instanceof FieldInfo || target instanceof MethodInfo || target instanceof MethodParameterInfo) {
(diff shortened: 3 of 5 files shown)


gsmet:
@n1hility could you have a look at this in priority? It introduced a big regression in both 1.13.5.Final and 2.0.0.CR1. I worked around it by adding a try/catch around the new HV code (but it's really a work around to avoid a plain build error) but we definitely need a fix in Jandex.

Also, given the issue is in the indexer, we will need a release of the Jandex Maven plugin which consumes this version.

Thanks!

yrodiere:
As @gsmet told me, the bug actually affected all type-uses, not just annotations on method parameter types.

I updated the fix, the tests and the PR description to address all type-uses.

famod:
> Also, given the issue is in the indexer, we will need a release of the Jandex Maven plugin which consumes this version.

FWIW, there was this idea to move Jandex to a separate org: https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/State.20of.20jandex-maven-plugin.3F
But it seems nothing has happened ever since and this fix here seems to important to wait for that, I guess.

n1hility:
Thanks @yrodiere ! I will take a look at this soon but skimming looks good. To get this into Quarkus I will need to backport since master is 3.x and that’s not fully baked. The way we test newer class file features than runtime  is a separate wildfly/typeannotation-test project. For 3.x the plan is to support 8 and later so will still need that approach for things like the proposed PR for module descriptor annotations (requires 9 or later). I can update the PR so not worry about having to refactor for that testing change

n1hility:
Fix is good, refactored test to externalize type annotations

n1hility (approved):

n1hility:
@gsmet @yrodiere Jandex 2.3.0.Final was released with this change to the jboss nexus repo. It will take ~24 hours to sync with central

yrodiere:
Thanks @n1hility!

Issue #124, opened 2021-07-27
Move to SmallRye


Ladicek:
For the 3.0 release, we want to move Jandex to SmallRye (and use the automated release process SmallRye has). Ideally, all Jandex repos (this one, the `typeannotation-test`, and the Maven plugin) would be brought into one (covered by #122).

The `groupId` would change from `org.jboss[.jandex]` to `io.smallrye` (or `io.smallrye.jandex`, but I'd probably prefer the first one :-) ). The `artifactId`s or packages wouldn't change (so packages would still be `org.jboss.jandex`), as that would be too big of a break.

darranl:
I don't think everyone will get the individual notifications from GitHub, maybe worth forwarding to the wildfly-dev list?

Ladicek:
I agree, but I also think we should do that after the 2.4.0.Final release. I mean -- what I just added to the 2.4.0.Final milestone was originally referred to as 3.0. Then we decided we should call it 2.4, as there's nothing breaking, and do a 3.0 with some breakages that are needed (including this one). I think that makes most sense, but until that is actually done, I'd like to leave the door open :-)

darranl:
+1 whenever makes more sense.

gastaldi:
@Ladicek moving to Smallrye sounds like a wonderful idea, +1 to that. About the package name, I wonder if that wouldn't be an opportunity to introduce some major API changes (eg. introduce a new API under `io.smallrye.jandex` and keep the original API under `org.jboss.jandex`)?

Ladicek:
I'm not sure we want to have 2 APIs. Also current implementation relies heavily on package-private methods, which precludes such split. I think we'll just live with some breaking changes :-)

Ladicek:
All the changes required in the code base (`groupId` and parent POM) are now done on the `smallrye` branch. I'm not closing this issue because Jandex is still under the `wildfly` org. I believe we should transfer this repo to the `smallrye` org when the 3.0 release is almost ready. In the meantime, the changes for the 3.0 release should be done as PRs against the `smallrye` branch.

Ladicek:
I just moved the Jandex repository to the `smallrye` GitHub org in preparation for a 3.0.0-RC1 release. Hence this is done for good.

jonathanlermitage:
If my understanding is correct, we can consider the `org.jboss.jandex` group-id moved to `io.smallrye.jandex`?  
[...]

PR #129, opened 2021-08-03
Read RuntimeInvisible*Annotations in Indexer

### LINKED ISSUE #120: Runtime Invisible Annotations
We had an issue a few weeks back smallrye/smallrye-open-api#716 requesting information to be utilized from a `@Retention(RetentionPolicy.CLASS)` annotation.  I'm curious if there is any interest in having runtime invisible annotations present in an index. I suppose the original reason for leaving them out was to more closely match reflection APIs, correct? Is this something that could be a candidate for version 3?

MikeEdgar:
Fixes #120 - I'm hoping they were not excluded deliberately originally :-) This also includes a refactoring of `processAttributes` to use a switch statement for readability.

### DIFF
--- a/core/src/main/java/org/jboss/jandex/IndexReaderV2.java
+++ b/core/src/main/java/org/jboss/jandex/IndexReaderV2.java
@@ -318,7 +318,10 @@ final class IndexReaderV2 extends IndexReaderImpl {
         DotName name = nameTable[stream.readPackedU32()];
         AnnotationTarget target = readAnnotationTarget(stream, caller);
         AnnotationValue[] values = readAnnotationValues(stream);
-        boolean visible = (version >= 11) && stream.readBoolean();
+        boolean visible = true;
+        if (version >= 11) {
+            visible = stream.readBoolean();
+        }
         return AnnotationInstance.create(name, visible, target, values);
     }
 
--- a/core/src/main/java/org/jboss/jandex/IndexReaderV2.java
+++ b/core/src/main/java/org/jboss/jandex/IndexReaderV2.java
@@ -48,7 +48,7 @@ import java.util.Set;
  */
 final class IndexReaderV2 extends IndexReaderImpl {
     static final int MIN_VERSION = 6;
-    static final int MAX_VERSION = 10;
+    static final int MAX_VERSION = 11;
     static final int MAX_DATA_VERSION = 4;
     private static final byte NULL_TARGET_TAG = 0;
     private static final byte FIELD_TAG = 1;
@@ -78,7 +78,8 @@ final class IndexReaderV2 extends IndexReaderImpl {
     private static final int HAS_ENCLOSING_METHOD = 1;
     private final static byte[] INIT_METHOD_NAME = Utils.toUTF8("<init>");
 
-    private PackedDataInputStream input;
+    private final PackedDataInputStream input;
+    private final int version;
     private byte[][] byteTable;
     private String[] stringTable;
     private DotName[] nameTable;
@@ -90,11 +91,12 @@ final class IndexReaderV2 extends IndexReaderImpl {
     private RecordComponentInternal[] recordComponentTable;
     private HashMap<DotName, Set<DotName>> users;
 
-    IndexReaderV2(PackedDataInputStream input) {
+    IndexReaderV2(PackedDataInputStream input, int version) {
         this.input = input;
+        this.version = version;
     }
 
-    Index read(int version) throws IOException {
+    Index read() throws IOException {
         try {
             PackedDataInputStream stream = this.input;
             int annotationsSize = stream.readPackedU32();
@@ -119,12 +121,12 @@ final class IndexReaderV2 extends IndexReaderImpl {
             if (version >= 10) {
                 readUsers(stream, usersSize);
             }
-            readMethodTable(stream, version);
+            readMethodTable(stream);
             readFieldTable(stream);
             if (version >= 10) {
                 readRecordComponentTable(stream);
             }
-            return readClasses(stream, annotationsSize, implementorsSize, subclassesSize, version);
+            return readClasses(stream, annotationsSize, implementorsSize, subclassesSize);
         } finally {
             byteTable = null;
             stringTable = null;
@@ -316,7 +318,8 @@ final class IndexReaderV2 extends IndexReaderImpl {
         DotName name = nameTable[stream.readPackedU32()];
         AnnotationTarget target = readAnnotationTarget(stream, caller);
         AnnotationValue[] values = readAnnotationValues(stream);
-        return new AnnotationInstance(name, target, values);
+        boolean visible = (version >= 11) && stream.readBoolean();
+        return AnnotationInstance.create(name, visible, target, values);
     }
 
     private Type[] readTypeListReference(PackedDataInputStream stream) throws IOException {
@@ -450,12 +453,12 @@ final class IndexReaderV2 extends IndexReaderImpl {
         throw new IllegalStateException("Invalid tag: " + tag);
     }
 
-    private void readMethodTable(PackedDataInputStream stream, int version) throws IOException {
+    private void readMethodTable(PackedDataInputStream stream) throws IOException {
         // Null holds the first slot
         int size = stream.readPackedU32() + 1;
         methodTable = new MethodInternal[size];
         for (int i = 1; i < size; i++) {
-            methodTable[i] = readMethodEntry(stream, version);
+            methodTable[i] = readMethodEntry(stream);
         }
 
     }
@@ -478,7 +481,7 @@ final class IndexReaderV2 extends IndexReaderImpl {
         }
     }
 
-    private MethodInternal readMethodEntry(PackedDataInputStream stream, int version) throws IOException {
+    private MethodInternal readMethodEntry(PackedDataInputStream stream) throws IOException {
         byte[] name = byteTable[stream.readPackedU32()];
         short flags = (short) stream.readPackedU32();
         Type[] typeParameters = typeListTable[stream.readPackedU32()];
@@ -538,7 +541,7 @@ final class IndexReaderV2 extends IndexReaderImpl {
     }
 
     private ClassInfo readClassEntry(PackedDataInputStream stream,
-            Map<DotName, List<AnnotationInstance>> masterAnnotations, int version) throws IOException {
+            Map<DotName, List<AnnotationInstance>> masterAnnotations) throws IOException {
         DotName name = nameTable[stream.readPackedU32()];
         short flags = (short) stream.readPackedU32();
         Type superType = typeTable[stream.readPackedU32()];
@@ -564,7 +567,7 @@ final class IndexReaderV2 extends IndexReaderImpl {
         if (hasNesting) {
             enclosingClass = nameTable[stream.readPackedU32()];
             simpleName = stringTable[stream.readPackedU32()];
-            enclosingMethod = hasEnclosingMethod ? readEnclosingMethod(stream, version) : null;
+            enclosingMethod = hasEnclosingMethod ? readEnclosingMethod(stream) : null;
         }
 
         int size = stream.readPackedU32();
@@ -784,7 +787,7 @@ final class IndexReaderV2 extends IndexReaderImpl {
         }
     }
 
-    private ClassInfo.EnclosingMethodInfo readEnclosingMethod(PackedDataInputStream stream, int version) throws IOException {
+    private ClassInfo.EnclosingMethodInfo readEnclosingMethod(PackedDataInputStream stream) throws IOException {
         if (version < 9 && stream.readUnsignedByte() != HAS_ENCLOSING_METHOD) {
             return null;
         }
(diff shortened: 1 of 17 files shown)


Ladicek:
My understanding is that this _was_ a deliberate decision :-) Perhaps also to make the index smaller. Not sure if these annotations should be accessible by default, or if opt-in should be required...

Side note: `case` clauses in `switch` should be indented :-)

MikeEdgar:
Thanks for the feedback. I fixed the indentation in the `switch`. I'll run a few tests to see what kind of memory impact this has to help inform whether an opt-in could help.

MikeEdgar:
I ran a test indexing the quarkus-app/lib jars for a "typical" Quarkus application with Resteasy, Hibernate, and some Camel libraries. The runtime-invisible annotations results in about an additional 1 MB of heap (56M before -> 57M). That could be mostly eliminated with an opt-in switch that could be used to selectively scan the invisible annotations, e.g. only for the application classes and not for dependencies.

Does that sound reasonable? If so, do you have any opinion on the form of the option/switch? Using a system property is my suggestion. Something like a boolean named `jandex.runtime-invisible-annotations`.

n1hility:
This is something we should definitely do, since we have had real asks for this. A great example is the Kotlin null annotations. That said, I have been unsure about how we handle the "surprise" of annotations showing up in Jandex but not via reflection, which could lead to framework errors.  There probably should be a flag on the annotation to mark whether or not it's visible, but do we need to go a step further and propvide mechanisms to filter (e.g. boolean includeInvisible)? If we filter on the top map API entry points (E.g. calls via Index.), what about people browsing from the class to the annotations.

My instinct is that its not worth API filtering, and we should just present them all with some flag to know the visibility. A good time to make this change would be in 3 since its a significant behavioral difference (new stuff showing up that didnt before).

Ladicek:
Yea I agree. `AnnotationInstance` could have a flag (`isRuntimeVisible` or something?), and I'd definitely consider this 3.0 material.

n1hility:
(BTW just for some background that's why it was not originally introduced in Jandex, there was no demand at the time, and the reflection inconsistency was a concern)

MikeEdgar:
One additional question - if this is for 3.0, does anyone still think there is a need to enable the invisible annotations via a system property, or some other mechanism like what is in this PR currently?

Ladicek:
Also, no, I don't think we need an opt-in anymore.

I originally thought it doesn't matter so much whether runtime-invisible annotations are present in the index or not. What does matter is the set of annotations returned from by the API. But if we put this into 3.0, it sounds palatable.

MikeEdgar:
I saw your comment in #124 and I'll rebase this and target the `smallrye` branch.

Ladicek:
Ah I didn't expect you to react so quickly :-) Well, maybe I can merge this as is then :-) I'll take a look. Thanks!

core/src/main/java/org/jboss/jandex/IndexReaderV2.java
```diff
@@ -316,7 +318,8 @@ private AnnotationInstance readAnnotationEntry(PackedDataInputStream stream, Ann
         DotName name = nameTable[stream.readPackedU32()];
         AnnotationTarget target = readAnnotationTarget(stream, caller);
         AnnotationValue[] values = readAnnotationValues(stream);
-        return new AnnotationInstance(name, target, values);
+        boolean visible = (version >= 11) && stream.readBoolean();
```
Ladicek, line 321: This will be `false` for `versions <= 10`, right? That would be wrong. Older indicies only contain runtime-visible annotations, so this should be `true` for `versions <= 10`. I'd write it like this:

```java
boolean visible = true;
if (version >= 11) {
    visible = stream.readBoolean();
}
```

Like this, it's identical to other places in the code that decide based on version.

core/src/main/java/org/jboss/jandex/Indexer.java
```diff
@@ -397,47 +437,128 @@ private void processRecordComponents(DataInputStream data) throws IOException {
...
+                processTypeAnnotations(data, target, annotationAttribute == HAS_RUNTIME_TYPE_ANNOTATION);
+                processed = true;
+                break;
+
+            case HAS_SIGNATURE:
                 processSignature(data, target);
-            } else if (annotationAttribute == HAS_EXCEPTIONS && target instanceof MethodInfo) {
-                processExceptions(data, (MethodInfo) target);
-            } else if (annotationAttribute == HAS_INNER_CLASSES && target instanceof ClassInfo) {
-                processInnerClasses(data, (ClassInfo) target);
```
Ladicek, line 420: To be fair, I find the old style (sequence of `if`s) much better to read than the new style (`switch` + `if` in each branch). The `processAttribute` method returning a `boolean` (and caller doing `skipFully` if `false`) is also not really necessary. The old style had `skipFully` as the last `else` branch, which is IMHO easier to understand than this.

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
```diff
@@ -468,6 +474,23 @@ public void testNullClass() throws IOException {
...
+            void execute(@RuntimeInvisible(placement = "arg") String arg) {
+            }
+        }
+
+        Index index = Index.of(Target.class);
+        DotName annoName = DotName.createSimple(RuntimeInvisible.class.getName());
+        List<AnnotationInstance> rtInvisible = index.getAnnotations(annoName);
+        assertEquals(2, rtInvisible.size());
+        assertEquals(0, Target.class.getDeclaredAnnotations().length);
+        assertEquals(0, Target.class.getDeclaredMethods()[0].getDeclaredAnnotations().length);
```
Ladicek, line 491: Pretty sure you want to assert against the `rtInvisible` data, not reflection data :-)

Ladicek:
Making `version` a field in readers/writers makes it impossible to reuse a single instance to read/write multiple versions, but we don't need that anywhere. And having `version` as field makes it unnecessary to pass it around as parameter, which is nice. So that's fine. Also the de-duplication of binary search of AnnotationInstance arrays is nice!

Out of the 3 comments I added, 2 of them are I believe logical errors that need to be fixed. The other is a matter of style, so... :-)

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
MikeEdgar: With this one, it was a sanity check that the two invisible annotations from the class and method param were invisible for reflection. Would it be better to remove that?

core/src/main/java/org/jboss/jandex/Indexer.java
MikeEdgar: I mainly did the switch to avoid duplicating the logic for the invisible annotation cases (the method param annotation case in particular). I'm fine to put it back to the other style if it's more readable for others, however. Let me know.

core/src/main/java/org/jboss/jandex/IndexReaderV2.java
MikeEdgar: Well, that's embarrassing... I'll fix it.

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
Ladicek: Ah gotcha. No that can stay, though I'd probably move it to the beginning, before creating the index. And I'd at least assert that the `AnnotationInstance`s correctly report that they're runtime-invisible.

core/src/main/java/org/jboss/jandex/Indexer.java
Ladicek: It should be straightforward to have an `if` condition matching more than 1 `annotationAttribute` value, I think?

MikeEdgar: No question about that :-). I'll update the PR later today.

core/src/main/java/org/jboss/jandex/IndexReaderV2.java
MikeEdgar: Fixed this. It turns out that the `IndexWriterV2` required some additional changes also to avoid writing the invisible annotations entirely for versions less than 11.

core/src/main/java/org/jboss/jandex/Indexer.java
MikeEdgar: Done

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
MikeEdgar: I think this should be good now. Adjusted the test to serialize/deserialize to version 11 and added another one for version 10.

core/src/main/java/org/jboss/jandex/Indexer.java
```diff
@@ -1904,6 +1951,14 @@ private boolean processConstantPool(DataInputStream stream) throws IOException {
...
                     } else if (len == RECORD_LEN && match(buf, offset, RECORD)) {
                         annoAttributes[pos] = HAS_RECORD;
+                    } else if (len == RUNTIME_INVISIBLE_ANNOTATIONS_LEN && match(buf, offset, RUNTIME_INVISIBLE_ANNOTATIONS)) {
+                        annoAttributes[pos] = HAS_RUNTIME_INVISIBLE_ANNOTATION;
+                    } else if (len == RUNTIME_INVISIBLE_PARAM_ANNOTATIONS_LEN
+                            && match(buf, offset, RUNTIME_INVISIBLE_PARAM_ANNOTATIONS)) {
+                        annoAttributes[pos] = HAS_RUNTIME_INVISIBLE_PARAM_ANNOTATION;
+                    } else if (len == RUNTIME_INVISIBLE_TYPE_ANNOTATIONS_LEN
+                            && match(buf, offset, RUNTIME_INVISIBLE_TYPE_ANNOTATIONS)) {
+                        annoAttributes[pos] = HAS_RECORD;
```
Ladicek, line 1961: Whoops :-)

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
```diff
@@ -468,6 +474,54 @@ public void testNullClass() throws IOException {
...
+    public void testRuntimeInvisiblePresentInV11() throws IOException {
+        @RuntimeInvisible(placement = "class")
+        class Target {
+            @SuppressWarnings("unused")
+            void execute(@RuntimeInvisible(placement = "arg") String arg) {
+            }
+        }
+
+        assertEquals(0, Target.class.getDeclaredAnnotations().length);
+        assertEquals(0, Target.class.getDeclaredMethods()[0].getDeclaredAnnotations().length);
```
Ladicek, line 487: The `@RuntimeInvisible` annotation is on a parameter, so shouldn't this be more like `Target.class.getDeclaredMethods()[0].getParameters()[0].getDeclaredAnnotations()`?

Ladicek: Also, could you add a runtime-invisible type annotation? E.g. `void execute(@RuntimeInvisible(placement = "arg") List<@RuntimeInvisible(placement = "type use") String> arg) {}`.

Ladicek:
Noticed one more mistake, which slipped through because a there wasn't a test. The test proposal I suggested should cover that.

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
MikeEdgar: Fixed and added a "type use" case

core/src/main/java/org/jboss/jandex/Indexer.java
MikeEdgar: I wish I could blame this on the rebase... fixed now.

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
MikeEdgar: Note that the test uses Java 8 features now - which I assume is fine for 3.0.

Ladicek: Yes, 3.0 will be based on 8, so that's fine.

core/src/test/java/org/jboss/jandex/test/BasicTestCase.java
```diff
@@ -475,39 +496,47 @@ public void testNullClass() throws IOException {
...
+                if (v == null) {
+                    v = new ArrayList<>();
+                }
+                v.add(a);
+                return v;
+            });
         }
+
+        assertEquals(1, placements.get("class").size());
+        // @RuntimeInvisible recorded on both method parameter and method parameter type
```
Ladicek, line 531: This is interesting, but correct I guess? Since the annotation has `ElementType.PARAMETER, ElementType.TYPE_USE`.

MikeEdgar: Yeah, exactly. I wasn't expecting it initially, but it makes sense.

Ladicek (approved):

Ladicek:
Looks good, thanks! I added one commit to your branch to somewhat simplify how `AnnotationInstance` is constructed. I changed the class to only have 1 `private` constructor, all callers now must use the `static create()` methods.

Please let me know if that looks OK to you, and I'll merge this.

MikeEdgar:
The `AnnotationInstance` changes look good to me - thanks @Ladicek

PR #130, opened 2021-08-09
Skip local name delimiter in last position

### LINKED ISSUE #97: Indexer vs DotName.equals() for a nested class with simple name that ends with $
If there is a nested class with simple name `Test$`, e.g. `org.jboss.jandex.test.IndexerTestCase$Test$` (note that although it's not recommended it's legal to use `$` in a class name) then `Indexer` decodes the `DotName` as:
```
componentized	true	
innerClass	true	
local	"" (id=108)	
prefix	DotName  (id=111)	
   componentized	true	
   innerClass	true	
   local	"Test" (id=117)
   prefix	DotName  (id=86)	
      componentized	true	
      innerClass	false	
      local	"test" (id=130)	
      prefix	DotName  (id=131)	
         ...
```
The `toString()` of the `DotName` is `org.jboss.jandex.test.IndexerTestCase$Test$` however the `equals()` method does not match the following:
```java
DotName org = DotName.createComponentized(null, "org");
DotName jboss = DotName.createComponentized(org, "jboss");
DotName jandex = DotName.createComponentized(jboss, "jandex");
DotName test = DotName.createComponentized(jandex, "test");
DotName indexerTestCase = DotName.createComponentized(test, "IndexerTestCase");
DotName testName = DotName.createComponentized(indexerTestCase, Test$.class.getSimpleName(), true);
```
whose `toString()` is also `org.jboss.jandex.test.IndexerTestCase$Test$`.

Maybe I'm missing something obvious.

MikeEdgar:
Fixes #97


Ladicek:
I see how this is conceptually a very simple solution, but it also _feels_ hacky (and scary, `NameTable.convertToName` is used from a couple of places that I don't really know).

I'll sleep over it, and see if I can make my peace with the idea :-)

MikeEdgar:
> `NameTable.convertToName` is used from a couple of places

The potential values for `delim` are `/` and `.`, plus the `$` specified in `convertToName(String, char)` itself. There are definitely more complex possibilities (i.e. a `$` in the middle of a class name). Consider the following name of a double nested class: `org.jboss.jandex.test.DotNameTestCase$Test$$InnerTest$`. Probably the only thing that can be known for certain is that the final `$` is not an inner class delimiter.

Ladicek:
Yea I agree about that. I was more thinking about `/` or `.`, but I have to take a look at those callers.

Ladicek:
OK, I debugged through this and it seems fine. It doesn't solve the problem of inner class names (and I believe it can't be solved in general without a major overhaul of the indexing algorithm, see https://github.com/wildfly/jandex/issues/97#issuecomment-897657547), but it does solve a problem of local parts of componentized `DotName`s being empty. Which should never happen, AFAICS.

Ladicek (approved):

PR #143, opened 2021-09-06
Use Locale.ROOT in String.format()

jorsol:
Is generally a bad practice to depend on the default platform locale, so this PR use Locale.ROOT for [String.format()](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#format-java.util.Locale-java.lang.String-java.lang.Object...-) calls.

### DIFF
--- a/src/main/java/org/jboss/jandex/Indexer.java
+++ b/src/main/java/org/jboss/jandex/Indexer.java
@@ -33,6 +33,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.IdentityHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import static org.jboss.jandex.ClassInfo.EnclosingMethodInfo;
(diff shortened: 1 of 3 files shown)


Ladicek:
I agree that depending on the platform locale is generally considered bad, but calling `String.format("my format string", my, parameters)` is super common and adding the locale argument just adds visual clutter. I'm not really sure we should do this.

jorsol:
Is the sad thing about Java defaults, but just because is super common doesn't mean it should be the correct option.

I agree that the locale argument adds visual clutter, but it potentially reduces the risk of the defaults:
* https://blog.thetaphi.de/2012/07/default-locales-default-charsets-and.html
* https://blog.joda.org/2012/12/annotating-jdk-default-data.html

So while in this case is not really critical as is just formating the output messages, it makes the output consistent.

Feel free to close it if the visual clutter is more annoying than the potential benefit.

Ladicek:
I slept on it and came to the conclusion that if we really want to use `String.format` as poor man's string interpolation, then we should just do it properly. Thanks for the PR!

Ladicek (approved):

PR #146, opened 2021-09-27
Add support for strings longer than 64 kB

### LINKED ISSUE #49: Exception when writing big indices
```
Caused by: java.io.UTFDataFormatException: encoded string too long: 68271 bytes
        at java.io.DataOutputStream.writeUTF(DataOutputStream.java:364)
        at java.io.DataOutputStream.writeUTF(DataOutputStream.java:323)
        at org.jboss.jandex.IndexWriterV2.writeStringTable(IndexWriterV2.java:216)
        at org.jboss.jandex.IndexWriterV2.write(IndexWriterV2.java:193)
        at org.jboss.jandex.IndexWriter.write(IndexWriter.java:107)
        at org.jboss.jandex.IndexWriter.write(IndexWriter.java:74)
```

Ladicek:
The `Data{Input,Output}Stream` classes only support [de]serializing
strings up to 64 kB of length, because they store/expect the length
of the string as a `short`. Sometimes, we need to read/write
longer strings. This commit does that by not using the built-in
support for strings, but doing a very similar thing manually.

A string is serialized as a number of bytes of an UTF-8 encoding
of the string (using our custom variable-length encoding of numbers),
followed by the byte array of the UTF-8 encoding of the string.
This only applies in a new persistent index format version. Older
versions are still handled using the old encoding.

Note that while `Data{Input,Output}Stream` uses a custom encoding
for strings, called "modified UTF-8", we use UTF-8 directly. For our
purposes, it shouldn't make a difference.

Fixes #49

### DIFF
--- a/core/src/main/java/org/jboss/jandex/IndexReaderV2.java
+++ b/core/src/main/java/org/jboss/jandex/IndexReaderV2.java
@@ -19,6 +19,7 @@
 package org.jboss.jandex;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -169,7 +170,14 @@ final class IndexReaderV2 extends IndexReaderImpl {
         int size = stream.readPackedU32() + 1;
         String[] stringTable = this.stringTable = new String[size];
         for (int i = 1; i < size; i++) {
-            stringTable[i] = stream.readUTF();
+            if (version >= 11) {
+                int len = stream.readPackedU32();
+                byte[] bytes = new byte[len];
+                stream.readFully(bytes, 0, len);
+                stringTable[i] = new String(bytes, StandardCharsets.UTF_8);
+            } else {
+                stringTable[i] = stream.readUTF();
+            }
         }
     }
 
--- a/core/src/main/java/org/jboss/jandex/IndexWriterV2.java
+++ b/core/src/main/java/org/jboss/jandex/IndexWriterV2.java
@@ -21,6 +21,7 @@ package org.jboss.jandex;
 import java.io.BufferedOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/jboss/jandex/IndexReaderV2.java
```diff
@@ -169,7 +170,14 @@ private void readStringTable(PackedDataInputStream stream) throws IOException {
         int size = stream.readPackedU32() + 1;
         String[] stringTable = this.stringTable = new String[size];
         for (int i = 1; i < size; i++) {
-            stringTable[i] = stream.readUTF();
+            if (version >= 11) {
+                int len = stream.readPackedU32();
+                byte[] bytes = new byte[len];
+                stream.readFully(bytes, 0, len);
+                stringTable[i] = new String(bytes, StandardCharsets.UTF_8);
```
n1hility, line 177: An optimization here could be to use InputStreamReader/OutputStreamWriter, so we avoid the large temporary array.

n1hility (changes requested):
One small quible. Feel free to override

core/src/main/java/org/jboss/jandex/IndexReaderV2.java
Ladicek: Good point, I'll check that.

Ladicek: Looking at it now and it actually wouldn't help. `OutputStreamWriter` writes through a `StreamEncoder`, which as the first thing copies the entire string to a `char[]`. Similarly, an `InputStreamReader` accepts a `char[]` which we'd have to provide, and constructing a `String` from it performs a copy.

So in the end, the code would be a little shorter, but would allocate the same temporary array _and then some_.

Ladicek:
This was a wrong solution, see https://github.com/wildfly/jandex/issues/49#issuecomment-932308817

Gonna submit another PR with a proper fix.

PR #149, opened 2021-10-02
[2.4] Support top-level class name with leading inner class delimiter

### LINKED ISSUE #148: Regression: Index containing class names that start with $ no longer have a consistent DotName.
CC @MikeEdgar 
This appears to be a regression with the parsing of DotNames caused by https://github.com/wildfly/jandex/commit/3779e75f7369ad4849f271b378fa2370f3d0e0f3

Consider the following test condition for the org.jboss.cdi.tck.tests.definition.name.$Dollar class:

``` java
Class<?> targetClass = $Dollar.class;
try (InputStream stream = targetClass.getClassLoader().getResourceAsStream(targetClass.getName().replace('.', '/') + ".class")) {
    indexer.index(stream);
    DotName name = DotName.createSimple(targetClass.getName());
    Assert.assertNotNull(index.getClassByName(name));
}
```

This assertion passes in 2.3.1, but fails in 2.4.0.
If I revert the above commit, the above assertion succeeds.

This is somewhat critical, as it causes a regression in the Jakarta TCK for CDI running against WildFly + Jandex 2.4.0 for the following test:
https://github.com/eclipse-ee4j/cdi-tck/blob/3.0/impl/src/main/java/org/jboss/cdi/tck/tests/definition/name/NameDefinitionTest.java#L109

```
[ERROR] org.jboss.cdi.tck.tests.definition.name.NameDefinitionTest.testNameStartingWithDollarCharacter  Time elapsed: 0.049 s  <<< FAILURE!
jakarta.enterprise.inject.UnsatisfiedResolutionException: Unable to resolve any beans of class org.jboss.cdi.tck.tests.definition.name.$Dollar
	at deployment.9d9c1e656827b7ae88a4f45c576e6894daaaf63.war//org.jboss.cdi.tck.tests.definition.name.NameDefinitionTest.testNameStartingWithDollarCharacter(NameDefinitionTest.java:110)
```

... and thus prevents WildFly from upgrading any of the smallrye components that require Jandex 2.4.0 (e.g. smallrye-open-api).

MikeEdgar:
Fixes #148 

Prior to #130 , top-level classes with names leading with `$` were parsed as nested classes contained within an empty-name class. Classes could still be retrieved from the index by simple `DotName`s as they were still equal via the `DotName#crossEquals` method.

If this is accepted I can open a separate PR for the `smallrye` branch.

### DIFF
--- a/src/main/java/org/jboss/jandex/NameTable.java
+++ b/src/main/java/org/jboss/jandex/NameTable.java
@@ -45,7 +45,7 @@ class NameTable {
         if (result != null)
             return result;
 
-        int loc = lastIndexOf(name, delim, '$');
+        int loc = lastIndexOf(name, delim);
         String local = intern(name.substring(loc + 1));
         DotName prefix = loc < 1 ? null : convertToName(intern(name.substring(0, loc)), delim);
         result = new DotName(prefix, local, true, loc > 0 && name.charAt(loc) == '$');
@@ -55,23 +55,21 @@ class NameTable {
         return result;
     }
 
-    private int lastIndexOf(String name, char delim1, char delim2) {
+    private int lastIndexOf(String name, char delim) {
         // Begin at second last position to avoid empty local name
         int pos = name.length() - 1;
         while (--pos >= 0) {
             char c = name.charAt(pos);
-
-            if (c == delim1) {
+            if (c == delim || c == '$') {
                 break;
             }
+        }
 
-            /*
-             * Only split on the inner class delimiter when it will not
-             * result in an empty string for the containing top-level class.
-             */
-            if (c == delim2 && pos > 0 && name.charAt(pos - 1) != delim1) {
-                break;
-            }
+        // avoid splitting on '$' if previous char is a delimiter or the '$'
+        // is in position 0, because subsequent split would produce an empty
+        // local name
+        if (pos >=0 && name.charAt(pos) == '$' && (pos == 0 || name.charAt(pos - 1) == delim)) {
+            pos--;
         }
 
         return pos;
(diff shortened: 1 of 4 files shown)


Ladicek:
Similarly to the previous fix for `$` at the _end_, couldn't we fix this case (`$` at the beginning) by stopping at the 2nd character, instead of going down to the 1st? Just curious :-)

MikeEdgar:
This is attempting to avoid creating a `DotName` with an empty local name for cases like `org/example/$MyClass`. It's intended to ignore the `$` when it's the first character of a top level class name. 

In light of the comments on #148, I would like to also review `DotName#equals`.

Ladicek:
Yea no, my comment above was stupid, I had to look at the code again in full. As long as we limit these changes to avoiding empty name components, we're fine. An empty name component is always a bug.

Ladicek:
I spent some more time looking into this and I think this PR is OK. Just for better readability, I think I would:

1. fold the `$` constant into the function. The `delim2` parameter is never anything else;
2. move the "if `$` and previous char is a delimiter" condition out of the loop.

Something like this:

```java
private int lastIndexOf(String name, char delim) {
    // Begin at second last position to avoid empty local name
    int pos = name.length() - 1;
    while (--pos >= 0) {
        char c = name.charAt(pos);
        if (c == delim || c == '$') {
            break;
        }
    }

    // avoid splitting on '$' if previous char is a delimiter,
    // because subsequent split would produce an empty local name
    if (pos > 0 && name.charAt(pos) == '$' && name.charAt(pos - 1) == delim) {
        pos--;
    }

    return pos;
}
```

But no pressure :-)

Ladicek:
Also, maybe we should add a test for a class (and/or package?) whose name _both_ starts and ends with a `$`.

Ladicek (approved):

Ladicek:
Could you please submit a PR for 3.0? Thanks!

PR #155, opened 2021-10-07
API Improvements and JDK update

stuartwdouglas:
- Update to JDK 1.8
- Add convenience methods to IndexView
I don't really know if this should go into a minor because of the JDK update, but this will get rid of a heap of boilerplate in Quarkus where we are wrapping String's in DotName for lookups.

### DIFF
--- a/core/src/main/java/org/jboss/jandex/IndexView.java
+++ b/core/src/main/java/org/jboss/jandex/IndexView.java
@@ -33,7 +33,7 @@ public interface IndexView {
      *
      * @return a collection of known classes
      */
-    public Collection<ClassInfo> getKnownClasses();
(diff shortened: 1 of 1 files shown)


Ladicek:
The `smallrye` branch is already updated to Java 1.8, and this will be released as 3.0. It certainly won't go into any 2.x release.

Adding the convenience methods certainly makes sense, but I _think_ it might be even better if we accepted `Class` objects? From what I saw in Quarkus, _that_ is the most common case.

stuartwdouglas:
I have added Class versions as well

Ladicek:
I rebased on the `smallrye` branch (which is Jandex 3.0 work in progress) and force-pushed here.

PR #168, opened 2022-01-14
Improve Index.of to also accept class files and JAR files

Ladicek:


MikeEdgar:
Any thoughts on a directory with jar files? Easily done by the client, but just for convenience.

Ladicek:
I didn't have a need for that yet :-) The things I'm adding here are things I've frequently used over past months.

MikeEdgar:
Just thinking of the situation "here's my app .jar and here's my lib/*.jar"

Ladicek:
It could be useful, but I think I'll let anyone _who actually has a use for it_ add it :-)

PR #173, opened 2022-02-14
introduce the notion of equivalence

### LINKED ISSUE #158: Add the notion of equivalence
Jandex classes often don't have `equals` at all, or have `equals` that isn't necessarily useful. That makes perfect sense e.g. in case of annotation overlaying, where you would have e.g. two `ClassInfo` objects for the same class, each having a different set of annotations. These `ClassInfo` objects _should not_ be considered _equal_, but they _should_ be considered _equivalent_.

I can see several ways how equivalence could be introduced to Jandex. I'm currently thinking it should be completely separate from the `AnnotationTarget` hierarchy. Something like this:

```java
public class EquivalenceKey {
    public static EquivalenceKey of(ClassInfo) { ... }
    public static EquivalenceKey of(MethodInfo) { ... }
    public static EquivalenceKey of(MethodParameterInfo) { ... }
    public static EquivalenceKey of(FieldInfo) { ... }

    // guaranteed to have `equals` and `hashCode` (and `toString`)
    // everything else would be private
}
```

Ladicek:
Resolves #158

### DIFF
--- a/core/src/main/java/org/jboss/jandex/EquivalenceKey.java
+++ b/core/src/main/java/org/jboss/jandex/EquivalenceKey.java
@@ -1,0 +1,637 @@
+package org.jboss.jandex;
+
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Establishes a notion of <em>equivalence</em> of Jandex objects. Two Jandex objects are equivalent if and only if
+ * they denote the same Java declaration or type, without taking into account any annotations. The prime use case
+ * is to assist with building annotation overlays on top of Jandex, where it is common to have multiple Jandex
+ * objects with different annotations, but otherwise equivalent.
+ * <p>
+ * In contrast, the common Jandex classes such as {@link ClassInfo}, {@link MethodInfo} or {@link FieldInfo}
+ * either don't provide equality at all (and hence can only be compared by identity) or provide strict equality,
+ * which includes presence or absence of annotations and comparison of their members, as well as other details.
+ * <p>
+ * An instance of this class, also called an <em>equivalence key</em>, provides 3 fundamental operations,
+ * corresponding to the 3 common methods all Java classes have:
+ * <ul>
+ * <li>{@code equals()}: when two Jandex objects are equivalent, their equivalence keys are equal</li>
+ * <li>{@code hashCode()}: consistent with {@code equals()} described above</li>
+ * <li>{@code toString()}: human readable representation of the equivalence key;
+ * format of the value is not guaranteed and may change without notice</li>
+ * </ul>
+ * In addition, equivalence keys are structured in an inheritance hierarchy that roughly corresponds
+ * to the inheritance hierarchy of Jandex objects. Therefore, the kind of the "source" Jandex object may be found
+ * by inspecting the class of the equivalence key:
+ * <ul>
+ * <li>{@code EquivalenceKey}
+ * <ul>
+ * <li>{@code DeclarationEquivalenceKey}
+ * <ul>
+ * <li>{@code ClassEquivalenceKey}</li>
+ * <li>{@code MethodEquivalenceKey}</li>
+ * <li>{@code MethodParameterEquivalenceKey}</li>
+ * <li>{@code FieldEquivalenceKey}</li>
+ * <li>{@code RecordComponentEquivalenceKey}</li>
+ * </ul>
+ * </li>
+ * <li>{@code TypeEquivalenceKey}
+ * <ul>
+ * <li>{@code ArrayTypeEquivalenceKey}</li>
+ * <li>{@code ClassTypeEquivalenceKey}</li>
+ * <li>{@code ParameterizedTypeEquivalenceKey}</li>
+ * <li>{@code PrimitiveTypeEquivalenceKey}</li>
+ * <li>{@code TypeVariableEquivalenceKey}</li>
+ * <li>{@code UnresolvedTypeVariableEquivalenceKey}</li>
+ * <li>{@code VoidTypeEquivalenceKey}</li>
+ * <li>{@code WildcardTypeEquivalenceKey}</li>
+ * </ul>
+ * </li>
+ * </ul>
+ * </li>
+ * </ul>
+ */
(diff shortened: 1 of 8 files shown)


Ladicek:
FYI @mkouba @FroMage, this should be pretty useful for anyone developing an annotation overlay on top of Jandex.

I'm not merging this yet, because I'd like to hear from you, and also because I really should add a few more tests :-)

core/src/main/java/org/jboss/jandex/WildcardType.java
```diff
@@ -26,7 +26,7 @@
  * @author Jason T. Greene
  */
 public class WildcardType extends Type {
-    private static Type OBJECT = new ClassType(DotName.OBJECT_NAME);
+    private static final Type OBJECT = new ClassType(DotName.OBJECT_NAME);
```
mkouba, line 29: OT: it might be useful to make this kind of constants public. WDYT? Other candidates are declared on the `org.jboss.jandex.DotName`, e.g. `DotName.OBJECT_NAME`.

Ladicek: Making some of the `DotName`-typed constants on `DotName` public certainly makes sense. Definitely not this one, which is on `WildcardType` and is of type `ClassType` :-)

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
```diff
@@ -0,0 +1,571 @@
...
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Where the common Jandex classes such as {@link ClassInfo}, {@link MethodInfo} or {@link FieldInfo}
+ * provide strict equality, which includes presence or absence of annotations and comparison of their members,
+ * this class provides <em>equivalence</em>. Two Jandex objects are equivalent if and only if they
+ * denote the same declaration or type, without taking into account any annotations. The prime use case
```
mkouba, line 12: > Where the common Jandex classes such as {link ClassInfo}, {link MethodInfo} or {link FieldInfo}
 provide strict equality, which includes presence or absence of annotations and comparison of their members,
this class provides <em>equivalence</em>. Two Jandex objects are equivalent if and only if they denote the same declaration or type, without taking into account any annotations.

Hm, I'm not sure I understand these sentences. For example the `ClassInfo` does not define `equals()`/`hashCode()` at all which means the object identity is used. For `FieldInfo` the name, flags, type and annotations are used. And for `MethodInfo` the name, flags, annotations, param names, param types, return type, default value (for annotations) and `receiverType` (I'm not sure what it means ;-). Maybe it wouldn't hurt to list all types of jandex objects and mention the "state" that is taken into account?

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
```diff
@@ -0,0 +1,571 @@
...
+ * </ul>
+ */
+public class EquivalenceKey {
+    /**
+     * Returns an equivalence key for given class.
+     *
+     * @param clazz the class, may be {@code null}
+     * @return equvalence key for given class, only {@code null} if {@code clazz == null}
+     */
+    public static ClassEquivalenceKey of(ClassInfo clazz) {
```
mkouba, line 31: Can we also have `public static EquivalenceKey of(AnnotationTarget target)`?

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
Ladicek: Fair enough, I'll rephrase this. Forgot about `ClassInfo` not providing equality at all.

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
Ladicek: That should be doable, yes. All cases should already be handled anyway.

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
```diff
@@ -0,0 +1,571 @@
...
+            MethodEquivalenceKey that = (MethodEquivalenceKey) o;
+            return Objects.equals(className, that.className)
+                    && Arrays.equals(methodName, that.methodName)
+                    && Arrays.equals(parameterTypes, that.parameterTypes)
+                    && Objects.equals(returnType, that.returnType);
+        }
+
+        @Override
+        public int hashCode() {
+            int result = Objects.hash(className, returnType);
```
mkouba, line 216: Perf experts always say that `Objects.hash()` is evil when used in hot paths because it allocates one more unnecessary array ;-). In fact, it also sometimes helps to compute the hashcode in the constructor and save it in a field. 

It depends on whether we expect the equivalence keys in hash maps and in hot paths.

core/src/main/java/org/jboss/jandex/WildcardType.java
mkouba: For sure, it should go to the `ClassType`. Ha, there is `org.jboss.jandex.ClassType.OBJECT_TYPE`!

FroMage (approved):
Looks useful, yes. But not really an annotation override ;)

FroMage:
Is it intentional that this is targeted at the `smallrye` branch and not `main`?

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
Ladicek: Rewrote.

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
Ladicek: Added.

core/src/main/java/org/jboss/jandex/WildcardType.java
Ladicek: Done.

core/src/main/java/org/jboss/jandex/EquivalenceKey.java
Ladicek: I'm ignoring this for now. If that ever becomes problematic, it's straightforward to improve.

Issue #175, opened 2022-03-07
IllegalStateException during indexing


attila-vegh:
I got the next exception from Jandex library:

Caused by: java.lang.IllegalStateException: Class not found in class table:scala.Function18$$anonfun$curried$1
    at org.jboss.jandex.IndexWriterV2.positionOf (IndexWriterV2.java:507)
    at org.jboss.jandex.IndexWriterV2.writeTypeEntry (IndexWriterV2.java:789)
    at org.jboss.jandex.IndexWriterV2.writeTypeTable (IndexWriterV2.java:271)
    at org.jboss.jandex.IndexWriterV2.write (IndexWriterV2.java:211)
    at org.jboss.jandex.IndexWriter.write (IndexWriter.java:110)
    at org.jboss.jandex.IndexWriter.write (IndexWriter.java:74)
    at org.jboss.jandex.JarIndexer.createJarIndex (JarIndexer.java:193)

Please investigate this issue. thanks

Ladicek:
I'd gladly investigate the issue, if you provided a way to reproduce. Since the stack trace starts at `JarIndexer.createJarIndex`, I assume you have a JAR file that you're trying to index. Is that JAR file available somewhere, or is it possible to build it from source that is available somewhere?

Ladicek:
The Scala library, at least the one I'm looking at (`org.scala:scala-library:2.13.8`) doesn't contain `scala.Function18$$anonfun$curried$1`. I'll need a way to reproduce the issue, I don't see a way around it.

Your screenshot is unfortunately useless. The `nameTable` in `IndexWriterV2` is a `HashMap` and `DotName` provides equality and hash code, so the fact that there are 2 different instances of `DotName` with the same content doesn't mean anything.

attila-vegh:
We are decided to filter out indexing scala jars. Now the indexing is ok

attila-vegh:
The Function is present in: scala-library-2.9.1.jar

Ladicek:
Thank you! Indeed running `java -jar core/target/jandex-3.0.0-SNAPSHOT.jar ~/Downloads/scala-library-2.9.1-1.jar` fails with the same exception:

```
$ java -jar core/target/jandex-3.0.0-SNAPSHOT.jar ~/Downloads/scala-library-2.9.1-1.jar 
java.lang.IllegalStateException: Class not found in class table:scala.collection.TraversableLike$$anonfun$isEmpty$1
[...]

PR #184, opened 2022-04-13
add IndexView.getKnownDirectSubinterfaces and getAllKnownSubinterfaces

### LINKED ISSUE #65: Add method to find extending interfaces of target interface
I would like to be able to (efficiently) look up all the interfaces that extends a target interface.
Say for example that I have 

```
public interface SomeInterface {

}
```

and there is an interface that extends is:

```
public interface SomeExtendingInterface extends SomeInterface {

}
```

 Neither `index.getAllKnownSubclasses()` nor `index.getAllKnownImplementors()` covers the aforementioned usecase so I am forced do the following (inefficient) lookup:

```
    List<ClassInfo> getAllInterfacesExtending(DotName target, IndexView index) {
        List<ClassInfo> result = new ArrayList<>();
        Collection<ClassInfo> knownClasses = index.getKnownClasses();
        for (ClassInfo clazz : knownClasses) {
            if (!Modifier.isInterface(clazz.flags())) {
                continue;
            }
            List<DotName> interfaceNames = clazz.interfaceNames();
            boolean found = false;
            for (DotName interfaceName : interfaceNames) {
                if (interfaceName.equals(target)) {
                    found = true;
                    break;
                }
            }
            if (found) {
                result.add(clazz);
            }
        }
        return result;
    }
```

Ladicek:
This is technically a breaking change, but only for users who
extend/implement the `IndexView` interface. Callers are fine.
It is tempting to fix `IndexView.getKnownDirectImplementors`
to no longer return subinterfaces, but we refrain from doing so
to maintain behavioral compatibility.

This change requires persistent format version bump. That has already
happened on the `smallrye` branch, so this commit doesn't bump
again, but note that this commit can't be backported to Jandex 2.x.

Resolves #65

### DIFF
--- a/core/src/main/java/org/jboss/jandex/IndexView.java
+++ b/core/src/main/java/org/jboss/jandex/IndexView.java
@@ -67,7 +67,7 @@ public interface IndexView {
     }
 
     /**
-     * Gets all known direct subclasses of the specified class name. A known direct
+     * Gets all known direct subclasses of the specified class. A known direct
      * subclass is one which was found during the scanning process; however, this is
      * often not the complete universe of subclasses, since typically indexes are
      * constructed per jar. It is expected that several indexes will need to be searched
@@ -83,7 +83,7 @@ public interface IndexView {
     Collection<ClassInfo> getKnownDirectSubclasses(DotName className);
 
     /**
-     * Gets all known direct subclasses of the specified class name. A known direct
+     * Gets all known direct subclasses of the specified class. A known direct
      * subclass is one which was found during the scanning process; however, this is
      * often not the complete universe of subclasses, since typically indexes are
      * constructed per jar. It is expected that several indexes will need to be searched
@@ -101,7 +101,7 @@ public interface IndexView {
     }
 
     /**
-     * Gets all known direct subclasses of the specified class name. A known direct
+     * Gets all known direct subclasses of the specified class. A known direct
      * subclass is one which was found during the scanning process; however, this is
      * often not the complete universe of subclasses, since typically indexes are
      * constructed per jar. It is expected that several indexes will need to be searched
@@ -119,7 +119,7 @@ public interface IndexView {
     }
 
     /**
-     * Returns all known (including non-direct) sub classes of the given class.
+     * Returns all known (including non-direct) subclasses of the given class.
      * I.e., returns all known classes that are assignable to the given class.
      *
      * @param className The class
@@ -129,7 +129,7 @@ public interface IndexView {
     Collection<ClassInfo> getAllKnownSubclasses(final DotName className);
 
     /**
-     * Returns all known (including non-direct) sub classes of the given class.
+     * Returns all known (including non-direct) subclasses of the given class.
      * I.e., returns all known classes that are assignable to the given class.
      *
      * @param className The class
@@ -141,7 +141,7 @@ public interface IndexView {
     }
 
     /**
-     * Returns all known (including non-direct) sub classes of the given class.
+     * Returns all known (including non-direct) subclasses of the given class.
      * I.e., returns all known classes that are assignable to the given class.
      *
      * @param clazz The class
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/jboss/jandex/IndexView.java
```diff
@@ -153,17 +153,107 @@ default Collection<ClassInfo> getAllKnownSubclasses(final Class<?> clazz) {
...
+     * environment (like an EE application server).
+     * <p>
+     * Note that this will only pick up direct subinterfaces of the interface. It will not
+     * pick up subinterfaces of subinterfaces.
+     *
+     * @param interfaceName the super interface of the desired subinterfaces
+     * @return a non-null list of all known subinterfaces of interfaceName
+     * @since 3.0
+     */
+    default Collection<ClassInfo> getKnownDirectSubinterfaces(String interfaceName) {
```
mkouba, line 187: Are these versions of the method really needed? I mean it can be useful yes, but all other methods accept `DotName` params so one could ask _"why string and class only for these two methods"_? Also clients may prefer these versions of methods which is suboptiomal because (1) componentized names are preferred and (2) it's always better to share `DotName` instances and avoid unnecessary allocations...

Ladicek: All other methods also have a `String` and `Class` overloads since https://github.com/wildfly/jandex/pull/155

When it comes to componentized `DotName`s, I think of it as a two-way street. When storing `DotName`s in a humongous data structure that must be as compact as possible (such as a Jandex index itself), using a componentized form is strongly preferred.

When it comes to one-off usages (such as these methods that lookup data from the index), it doesn't really matter in my opinion. If you share `DotName` instances (typically by storing them into `static final` fields), you prevent unnecessary allocations at the expense of higher old gen usage. If you create them on demand, you cause higher allocation rate. In this particular case, the freshly created `DotName` instances are never stored anywhere, so they die very young, and collecting a new gen is rather cheap.

So I don't really think this is an issue.

mkouba: > All other methods also have a `String` and `Class` overloads since #155

Ok, didn't know about that one.

> 
> When it comes to componentized `DotName`s, I think of it as a two-way street. When storing `DotName`s in a humongous data structure that must be as compact as possible (such as a Jandex index itself), using a componentized form is strongly preferred.

AFAIK the componentized version also has a bit more efficient `equals()`...

> 
> When it comes to one-off usages (such as these methods that lookup data from the index), it doesn't really matter in my opinion. If you share `DotName` instances (typically by storing them into `static final` fields), you prevent unnecessary allocations at the expense of higher old gen usage. If you create them on demand, you cause higher allocation rate. In this particular case, the freshly created `DotName` instances are never stored anywhere, so they die very young, and collecting a new gen is rather cheap.

Ok, I won't argue ;-) although I think that these methods make the API less readable and the _"collecting a new gen is rather cheap"_ argumet is a bit disputable.

Ladicek: Actually I had to fix `DotName.equals()` for the case when both `DotName`s are componentized, because the previous implementation assumed that componentization is canonical, which it is not (see #175 and #176 if you wanna know more, but beware, it's a long brain dump of mine :-) ). So `DotName.equals()` is actually most efficient when both `DotName`s are _not_ componentized, because that's just a string comparison.

Performance disputes are meaningless without measurements, and I don't have any -- my only argument is that Stuart clearly thought it's fine in #155 :-)

mkouba: > Actually I had to fix `DotName.equals()` for the case when both `DotName`s are componentized, because the previous implementation assumed that componentization is canonical, which it is not (see #175 and #176 if you wanna know more, but beware, it's a long brain dump of mine :-) ). So `DotName.equals()` is actually most efficient when both `DotName`s are _not_ componentized, because that's just a string comparison.

Interesting! 

> 
> Performance disputes are meaningless without measurements, and I don't have any -- my only argument is that Stuart clearly thought it's fine in #155 :-)

+1 :D

PR #192, opened 2022-04-26
uniform handling of implicitly declared (aka mandated) and synthetic parameters

### LINKED ISSUE #45: Support a reliable way to get a parameter type

Ladicek:
Resolves #45

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,6 @@ jobs:
         os:
         - ubuntu-latest
         - windows-latest
-        - macos-latest
         exclude:
         - java: 8
           compiler: ecj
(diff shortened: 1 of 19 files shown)


Ladicek:
Draft for now, because this PR is a bit inefficient in how methods are serialized. Most of the time, both parameter arrays (signature-based and descriptor-based) are identical, in which case, we really should only serialize one. Need to work on that.

Ladicek:
Actually my comment above is wrong. Type arrays are interned and the intern pool is serialized once. When serializing a method, we only write down an _index_ into the intern pool.

We _could_ still check whether the 2 parameter arrays are different, and if they are identical, we'd just write down a boolean flag. If they are different, we'd write down a boolean flag and the index into the intern pool. That makes a difference of roughly 50 kB when indexing OpenJDK 8 `rt.jar`, which is IMHO negligible (the entire index has roughly 6 MB).

PR #203, opened 2022-05-17
add methods to Index to work with packages

### LINKED ISSUE #96: Get all classes from a package
I know Jandex is mostly about annotations but I was wondering if we could extend it a bit to keep an index of packages, subpackages and classes in these packages.

When building native applications with external libraries, it can happen that all the content of a package needs to be registered for reflection and that it is the only way to really be sure we do not forget anything and future-proof enough.

Listing classes manually is very cumbersome and brittle for large APIs.

Given the index is a closed world, it sounds feasible, right?

Thoughts?

Ladicek:
Two new methods, `getClassesInPackage` and `getSubpackages`, are added
to `IndexView`. Unlike other index methods, these don't expect to have
precomputed answers stored in the persistent format. The packages info
can be recomputed rather easily from the full set of known classes, so
these methods just initialize their "index" on demand. First call then
may be slow, but subsequent calls will be fast.

Resolves #96

### DIFF
--- a/core/src/main/java/org/jboss/jandex/CompositeIndex.java
+++ b/core/src/main/java/org/jboss/jandex/CompositeIndex.java
@@ -320,4 +320,33 @@ public class CompositeIndex implements IndexView {
         }
         return Collections.unmodifiableCollection(users);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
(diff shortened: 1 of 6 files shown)


Ladicek:
This is API breaking, because it adds 2 new methods to `IndexView`.

PR #221, opened 2022-06-20
Add proper support for recursive type parameters

### LINKED ISSUE #88: Generics - recursive bound results in UnresolvedTypeVariable
If a class declares a recursive bound (which should be legal and often used e.g. in builders) such as `interface Score<S extends Score<S>> {}` the type hierarchy contains an `UnresolvedTypeVariable`. And according to the javadoc _"This type will only occur as a result of a bug, or a non-compliant Java class file"_. 

A test can be found here:
https://github.com/mkouba/jandex/commit/2141db3e375f824a486dc909d00b2f30673017a9

In fact, I'm not quite sure whether it's a bug or how to handle this correctly.

Ladicek:
This commit adds a new kind of type, a `TypeVariableReference`, which is used
in a class/method list of type parameters to represent a type variable that
isn't fully defined yet. For simple recursive type parameters, such as
`T extends Comparable<T>`, this naturally means that type variables occurring
in their own definitions are represented as references. For mutually recursive
type parameters, type variables that occur in their definitions or before their
definitions are represented as references. A type variable reference can be
`follow()`-ed to obtain the underlying type variable.

The `TypeVariableReference` is mutable, initially points to `null` and must
later be patched. This occurs on several occasions:

1. When indexing a class, generic signatures are parsed, which includes
   resolution of type variables (where unresolved type variables that are
   found recursive are replaced by references) and then patching of newly
   introduced references.
2. Whe completing an index, type annotations are propagated, which includes
   creating deep copies of type variables that include references. To create
   such deep copies, new references are introduced, to avoid sharing. Such
   new references are later patched.
3. When deserializing an index, types are read in topological order (if a type
   is composite, it is read after its constituent types), which also requires
   patching references (because they are read earlier).

This change requires persistent format version bump. That has already
happened on the `smallrye` branch, so this commit doesn't bump again,
but note that this commit can't be backported to Jandex 2.x.

Resolves #88

### DIFF
--- a/core/src/main/java/org/jboss/jandex/EquivalenceKey.java
+++ b/core/src/main/java/org/jboss/jandex/EquivalenceKey.java
@@ -47,6 +47,7 @@ import java.util.StringJoiner;
  * <li>{@code ParameterizedTypeEquivalenceKey}</li>
  * <li>{@code PrimitiveTypeEquivalenceKey}</li>
  * <li>{@code TypeVariableEquivalenceKey}</li>
+ * <li>{@code TypeVariableReferenceEquivalenceKey}</li>
  * <li>{@code UnresolvedTypeVariableEquivalenceKey}</li>
  * <li>{@code VoidTypeEquivalenceKey}</li>
  * <li>{@code WildcardTypeEquivalenceKey}</li>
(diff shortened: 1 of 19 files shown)


Ladicek:
Added tests for parameterized type with owner, and that should be all.

PR #259, opened 2022-09-09
Add methods to access enum constants

Ladicek:


Ladicek:
I added this as a draft to run CI with various compilers. All runs pass, so all tested compilers store the fields in bytecode in declaration order.

PR #286, opened 2022-11-23
Reproducible serialisation, fixes #213

### LINKED ISSUE #213: Reconsider reproducible serialization
There were attempts to make Jandex reproducible before:
- https://github.com/wildfly/jandex/issues/71
- https://github.com/wildfly/jandex-maven-plugin/pull/35

So far, this has been dismissed, but we may want to reconsider. One reason would be to allow incremental builds. If a Maven module carries a Jandex index, rebuilding always changes the JAR, even though the classes perhaps didn't change at all -- the only thing that changed was the index. (Immediate question: would it be possible to avoid rebuilding the index if the classes weren't rebuilt?)

This is not a 3.0 topic, but may be a 3.x topic. This issue is mainly for collecting feedback.

gnodet:


gnodet:
> Can you please comment on what you want to achieve with this? I'd like to note that reindexing the same set of files is very likely to lead to a different index, so this is unlikely to help with reproducibility (#213).

I can understand that quarks has problems with reproducible builds, but in this case, we're using gander outside quarkus.  Topological order is fine and should be stable to lead to a reproducible index.

gnodet:
I'll work on additional commits to keep the index stable.

gnodet:
Fwiw, jandex 1.x was writting reproducible indices.  This PR restores that and avoid rewriting the index file in case it has not changed.

gnodet:
I'll investigate the test which fails on JDK 20, but first, let me know if you would merge this PR, else there's no point in spending time in investigating...

Ladicek:
OK so if all the changes needed are the `PrimitiveType` hash code, then that seems sensible. I wouldn't want to merge such invasive changes as #72.

I would also be fine with the solution for detecting unchanged files using hash trees, as outlined in #213, but that would require bumping the index version -- so if this PR achieves reproducibility, that's even better.

gnodet:
> OK so if all the changes needed are the `PrimitiveType` hash code, then that seems sensible. I wouldn't want to merge such invasive changes as #72.

Sounds good.  I can check with a larger code base if the index serialisation is reproducible if needed.
 
> I would also be fine with the solution for detecting unchanged files using hash trees, as outlined in #213, but that would require bumping the index version -- so if this PR achieves reproducibility, that's even better.

The computation of the index is fast enough, I don't think we really need to detect unchanged files.  Not rewriting the output looks sufficient and will avoid downstream projects in reactor builds to be recompiled (because jars would have changed).

Ladicek:
I agree -- the way I would detect unchanged files would still require building the entire index in memory. It would just skip writing it to the file if no input changed. If this PR achieve the same, I'm happy with that.

Ladicek:
I'm merging this with an explicit statement that I do _not_ consider index reproducibility an intentional feature that's subject to the compatibility promise. It's a best effort kind of thing that I will not hesitate breaking if required.

That said -- thanks a lot for the PR!

PR #293, opened 2023-01-10
Add API for reconstructing descriptors and generic signatures

Ladicek:
This API includes type variable substitution, which is used on several places in Quarkus. If the substitution function returns a type for some type variable identifier, the reconstructed descriptor/signature contains a descriptor/signature of the provided type instead of the original type variable.


Ladicek:
Draft for now, as I need to test if the changes here are sufficient for Quarkus usage of the type variable substitution feature.

PR #299, opened 2023-02-28
Build generates reproducible jars

bdemers:
Per: https://maven.apache.org/guides/mini/guide-reproducible-builds.html

* Updates maven-bundle-plugin to 5.1.8 (which removes the `Bnd-LastModified` field from the manifest)
* Adds `project.build.outputTimestamp` to the root pom,  the release plugin will update this value when preparing a release

NOTE: The `test-data` build is NOT reproducible as the jar is patched out of band using the `jar` command


Ladicek (approved):
I don't think we ever intended for the build to be reproducible, but if this is all it takes, then I have no problem accepting that :-)

PR #315, opened 2023-06-12
support directory in jar

### LINKED ISSUE #314: classloader.getResource(entryName) return null
In the `jar-index` feature
 classloader method:` classloader.getResource(entryName)` when the entryName is a drectory ,this method would return null. In some opensource tools use this feature to load the resouces in the same folder would fail.

wangshihui:
support directory in jar,  classloader method: classloader.getResource(entryName) when the entryName is a drectory ,this method would return null.In some opensource tools use this feature to load the resouces in the same folder 
would faile

Resolves #314

### DIFF
--- a/maven-plugin/src/it/jar/verify.groovy
+++ b/maven-plugin/src/it/jar/verify.groovy
@@ -6,6 +6,7 @@ assert jarFile.exists() : "File ${jarFile} does not exist"
 assert jarFile.length() > 0 : "File ${jarFile} is empty"
 
 def jar = new ZipFile(jarFile)
+assert jar.getEntry("META-INF/") != null : "JAR ${jarFile} doesn't contain a META-INF directory"
 def indexEntry = jar.getEntry("META-INF/jandex.idx")
 assert indexEntry != null : "JAR ${jarFile} doesn't contain an index"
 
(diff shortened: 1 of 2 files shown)


Ladicek:
Ah so this is to retain the directory structure in a JAR that is being reindexed. That seems sensible, but I'm not sure about the code that follows the `if` that's modified here -- does it handle directory entries correctly? I'd expect problems there.

wangshihui:
> Ah so this is to retain the directory structure in a JAR that is being reindexed. That seems sensible, but I'm not sure about the code that follows the `if` that's modified here -- does it handle directory entries correctly? I'd expect problems there.

I test in my project after modification，every thing works ok.  and we cuurently store and use it in our private repository. 


if  **exlude** the dirctory structure from the **jar**,  classloader.getResource("")  would retrun null. so if application want to  get all resources from a  named directory is complexed. 
eg： get the all properties from  `classpath:config`, should  walk through all the entries. then filer the prefix with ` config`

also: a lot of opensource tools use this feature of  ` classloader.getResources(dirctory)`, if no direcotry enty in the jar file，these tools can not work 。

Ladicek (approved):

PR #320, opened 2023-08-11
Allow configuring persistent index format version in the Maven plugin

### LINKED ISSUE #137: Configurable persistent index format version in the Maven plugin
The Maven plugin could expose a configuration property for setting which index version to write. It would then call `IndexWriter.write(Index, int)` instead of just `IndexWriter.write(Index)`. Default would be unconfigured, which would lead to calling `IndexWriter.write(Index)` and hence using the latest index version.

The purpose would be to decouple Jandex version bump from Jandex persistent index format version bump. Libraries that produce and distribute a Jandex index could decide to bump Jandex without bumping the persistent index format, so that the index remains readable by older Jandex versions.

On the other hand, there's clear desire to make libraries ship latest indices, so that new Jandex features can be used, so maybe we would only allow configuring a _subset_ of possible index versions? Say, current and current - 1, something like that.

iam-afk:
Resolves #137

### DIFF
--- a/maven-plugin/src/it/indexVersion/pom.xml
+++ b/maven-plugin/src/it/indexVersion/pom.xml
@@ -4,7 +4,7 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>org.jboss.jandex</groupId>
-  <artifactId>jandex-maven-plugin-smoketest</artifactId>
+  <artifactId>jandex-maven-plugin-indexVersion</artifactId>
   <version>1.0-SNAPSHOT</version>
 
   <properties>
@@ -33,9 +33,9 @@
             <goals>
               <goal>jandex</goal>
             </goals>
-            <!-- phase is 'process-classes by default' -->
             <configuration>
-              <!-- Nothing needed here for simple cases -->
+              <!-- oldest version of V2 index -->
+              <indexVersion>6</indexVersion>
             </configuration>
           </execution>
         </executions>
--- a/maven-plugin/src/it/indexVersion/src/main/java/org/jboss/jandex/maven/indexVersion/SomeClass.java
+++ b/maven-plugin/src/it/indexVersion/src/main/java/org/jboss/jandex/maven/indexVersion/SomeClass.java
@@ -1,0 +1,4 @@
+package org.jboss.jandex.maven.indexVersion;
+
+public class SomeClass {
+}
--- a/maven-plugin/src/it/indexVersion/verify.groovy
+++ b/maven-plugin/src/it/indexVersion/verify.groovy
@@ -1,0 +1,13 @@
+import org.jboss.jandex.IndexReader
+import org.jboss.jandex.PackedDataInputStream
+
+def jandexFile = new File(basedir, 'target/classes/META-INF/jandex.idx')
+assert jandexFile.exists() : "File ${jandexFile} does not exist"
+assert jandexFile.length() > 0 : "File ${jandexFile} is empty"
+
+def data = new PackedDataInputStream(jandexFile.newInputStream())
+assert data.readInt() == 0xBABE1F15 as int // magic
+assert data.readUnsignedByte() == 6 // version
+
+def index = new IndexReader(jandexFile.newInputStream()).read()
+assert index.getKnownClasses().size() == 1 : "Index ${jandexFile} does not contain exactly 1 class"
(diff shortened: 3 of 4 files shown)


Ladicek (approved):

Ladicek:
Out of curiosity, is this something you actually need, or did you implement it just "for fun"?

iam-afk:
😃 In short the change did not help me to solve my original issue. 

We have a project where we use Jandex 1 or 2 (because of Hibernate depends on Jandex and Wildfly provide Jandex module). But once jandex-maven-plugin stopped working with latest JDK ea builds (5f044c136129c4d8b57188fea4f01f76f1d4ea88). I thought I could use latest jandex-maven-plugin 3 to create persistent index compatible with Jandex 2. But despite the fact that the versions of the formats have become the same, Jandex 2 could not read the index (unknown annotation tag value, if I remember it right)...

Since there is the issue #137 (someone else need something like this) and I have already made changes, why not make a PR.

Ladicek:
OK, that is interesting. If you can provide a reproducer, I can take a look.

In the meantime, I'll try to find some time to write a test for this and then merge.

iam-afk:
> OK, that is interesting. If you can provide a reproducer, I can take a look.
> 
> In the meantime, I'll try to find some time to write a test for this and then merge.

I was able to reproduce a problem (or maybe one case) with this example:
```java
import java.util.List;

public interface X<T extends List<T>> { }
```

```shell
javac --release 11 X.java
```

Test runner `scratch.java`:
```java
import org.jboss.jandex.Index;
import org.jboss.jandex.IndexReader;
import org.jboss.jandex.IndexWriter;
import org.jboss.jandex.Indexer;

import java.io.IOException;
import java.io.InputStream;
import java.io.OutputStream;
import java.nio.file.Files;
import java.nio.file.Paths;

class Scratch {
    public static void main(String[] args) throws IOException {
        if (args.length > 0 && "create".equals(args[0])) {
            try (InputStream inputStream = Files.newInputStream(Paths.get("X.class"))) {
                indexer.index(inputStream);
            }

            try (OutputStream outputStream = Files.newOutputStream(Paths.get("jandex.idx"))) {
                IndexWriter indexWriter = new IndexWriter(outputStream);
                indexWriter.write(index, 9);
            }
        } else {
            try (InputStream inputStream = Files.newInputStream(Paths.get("jandex.idx"))) {
                Index index = new IndexReader(inputStream).read();
                System.out.println("index.getKnownClasses().size() = " + index.getKnownClasses().size());
            }
        }
    }
}
```

Create index with Jandex 3, version `9` (max for version `2.2.3`):
```shell
java -cp jandex-3.1.2.jar --source 11 scratch.java create
```
Try to read index with Jandex 2:
```shell
java -cp jandex-2.2.3.Final.jar --source 11 scratch.java
```
Get error:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 1
	at org.jboss.jandex.IndexReaderV2.readAnnotations(IndexReaderV2.java:208)
	at org.jboss.jandex.IndexReaderV2.readTypeEntry(IndexReaderV2.java:341)
	at org.jboss.jandex.IndexReaderV2.readTypeTable(IndexReaderV2.java:175)
	at org.jboss.jandex.IndexReaderV2.read(IndexReaderV2.java:109)
	at org.jboss.jandex.IndexReader.read(IndexReader.java:76)
	at Scratch.main(scratch.java:27)
```

But this works fine:
```shell
java -cp jandex-3.1.2.jar --source 11 scratch.java
```
```
index.getKnownClasses().size() = 1
```

Ladicek:
I'll take a look. Apparently we broke compatibility, which is bad. Thanks for the reproducer!

Ladicek:
OK, the problem is that Jandex 3 added a new kind of types, type variable references, and the serialization code doesn't account for that when serializing to an older version. Fortunately, there's a simple fix -- what Jandex 3 represents as a type variable reference, Jandex 2 represented as an unresolved type variable, so we'll just resort to writing an unresolved type variable for older versions of the serialization format. Here's a PR: https://github.com/smallrye/jandex/pull/324 If you could build Jandex from that branch and check, that would be great. It passes your reproducer above.

iam-afk:
Thanks, I'll try after the weekend.

iam-afk:
> OK, the problem is that Jandex 3 added a new kind of types, type variable references, and the serialization code doesn't account for that when serializing to an older version. Fortunately, there's a simple fix -- what Jandex 3 represents as a type variable reference, Jandex 2 represented as an unresolved type variable, so we'll just resort to writing an unresolved type variable for older versions of the serialization format. Here's a PR: #324 If you could build Jandex from that branch and check, that would be great. It passes your reproducer above.

Yes, now it works on the whole project! Thanks! Wait for release...

PR #323, opened 2023-08-24
Fixes #322 by reverting internHashCode()

### LINKED ISSUE #322: Indexing takes over 100x longer on big file
When indexing a big file that contains around 16000 generated classes, the indexing process takes around 100 more time.

Jandex 3.1.0  ~ 5 seconds
Jandex 3.1.1  ~ 490 seconds

reinhapa:
Fixes performance degradation caused by changing the `internHashCode`


Ladicek:
We'll have to figure out something better, because this unfortunately breaks reproducibility.

reinhapa:
@Ladicek I'm open to any fix that restores the old performance...

Ladicek:
I hear you, and I'll have to think more about it. This is arguably a corner case, but such performance loss is certainly not acceptable.

PR #330, opened 2023-09-22
release 3.1.4

Ladicek:


Ladicek:
The artifacts ended up in Maven Central, but the release job failed. Therefore, the release commits were not pushed, there is no tag, the milestone wasn't closed.

I consider this release invalid and I'm attempting 3.1.5.

PR #344, opened 2023-11-27
Update plugin dependencies

jorsol:
This updates the plugins used in the project, particularly the plexus-utils update is important as it fixes a false difference detected with CachingOutputStream/CachingWriter when streams are flushed.

### DIFF
--- a/maven-plugin/pom.xml
+++ b/maven-plugin/pom.xml
@@ -72,6 +72,22 @@
                         <artifactId>jandex</artifactId>
                         <version>${project.version}</version>
                     </dependency>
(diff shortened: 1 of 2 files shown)


Ladicek (approved):

jorsol:
> Too bad I released 3.1.6 just a few hours ago :-/

Yeap, I was not quick enough :(

Ladicek:
Maven Invoker tests were failing due to https://issues.apache.org/jira/browse/GROOVY-11197, but fortunately, Groovy 4.0.16 was released yesterday. So I added an override of the Groovy version in the Maven Invoker plugin.

PR #346, opened 2023-12-11
Add empty and stacked indexes

### LINKED ISSUE #142: Add a "stacked" composite index
Jandex includes `CompositeIndex`, which combines multiple indices that don't overlap. When the indices do overlap, `CompositeIndex` will return duplicates.

We could add an `OverlayIndex` which would overlay one index on top of another (or perhaps a stack of indices, not just two). Some methods on `IndexView` are perhaps hard to define, but the most common ones should be possible.

Something like this:

```java
public class OverlayIndex implements IndexView {
    private final List<IndexView> indices;

    Collection<ClassInfo> getKnownClasses() {
        // call `getKnownClasses` on each index, in order, keep track of already seen class names
        // and skip a class if its name was already seen
    }

    ClassInfo getClassByName(DotName className) {
        // iterate through `indices` and return the first found
    }

    Collection<ClassInfo> getKnownDirectSubclasses(DotName className) {
        // don't know here? perhaps call `getKnownDirectSubclasses` on each index, in order,
        // keep track of already seen class names and skip a class if its name was already seen
    }

    Collection<ClassInfo> getAllKnownSubclasses(DotName className) {
        // don't know here? perhaps call `getAllKnownSubclasses` on each index, in order,
        // keep track of already seen class names and skip a class if its name was already seen
    }

    Collection<ClassInfo> getKnownDirectImplementors(DotName className) {
        // don't know here? perhaps call `getKnownDirectImplementors` on each index, in order,
        // keep track of already seen class names and skip a class if its name was already seen
    }

    Collection<ClassInfo> getAllKnownImplementors(final DotName interfaceName) {
        // don't know here? perhaps call `getAllKnownImplementors` on each index, in order,
        // keep track of already seen class names and skip a class if its name was already seen
    }

    Collection<AnnotationInstance> getAnnotations(DotName annotationName) {
        // call `getAnnotations` on each index, keep track of already seen `AnnotationTarget`s
        // and skip an `AnnotationTarget` if it was already seen
        //
        // note that `AnnotationTarget` doesn't have well defined equality, so this may be harder than it seems
    }

    Collection<AnnotationInstance> getAnnotationsWithRepeatable(DotName annotationName, IndexView index) {
        // same as getAnnotations
    }

    Collection<ModuleInfo> getKnownModules() {
        // same as `getKnownClasses` probably?
    }

    ModuleInfo getModuleByName(DotName moduleName) {
        // same as `getClassByName` probably?
    }

    Collection<ClassInfo> getKnownUsers(DotName className) {
        // call `getKnownUsers` on each index, in order, keep track of already seen class names
        // and skip a class if its name was already seen
    }
}
```

Ladicek:
Fixes #142

### DIFF
--- a/core/src/main/java/org/jboss/jandex/EmptyIndex.java
+++ b/core/src/main/java/org/jboss/jandex/EmptyIndex.java
@@ -1,0 +1,93 @@
+package org.jboss.jandex;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Immutable empty index; that is, an index that doesn't contain any class. All methods return either
+ * an empty collection, or {@code null}.
+ *
+ * @since 3.2.0
+ */
+public final class EmptyIndex implements IndexView {
+    public static final EmptyIndex INSTANCE = new EmptyIndex();
+
+    private EmptyIndex() {
+    }
+
+    @Override
+    public Collection<ClassInfo> getKnownClasses() {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public ClassInfo getClassByName(DotName className) {
+        return null;
+    }
+
+    @Override
+    public Collection<ClassInfo> getKnownDirectSubclasses(DotName className) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<ClassInfo> getAllKnownSubclasses(DotName className) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<ClassInfo> getKnownDirectSubinterfaces(DotName interfaceName) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<ClassInfo> getAllKnownSubinterfaces(DotName interfaceName) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<ClassInfo> getKnownDirectImplementors(DotName interfaceName) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<ClassInfo> getAllKnownImplementors(DotName interfaceName) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<AnnotationInstance> getAnnotations(DotName annotationName) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<AnnotationInstance> getAnnotationsWithRepeatable(DotName annotationName, IndexView index) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<ModuleInfo> getKnownModules() {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public ModuleInfo getModuleByName(DotName moduleName) {
+        return null;
+    }
+
+    @Override
+    public Collection<ClassInfo> getKnownUsers(DotName className) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Collection<ClassInfo> getClassesInPackage(DotName packageName) {
+        return Collections.emptySet();
+    }
+
+    @Override
+    public Set<DotName> getSubpackages(DotName packageName) {
+        return Collections.emptySet();
+    }
+}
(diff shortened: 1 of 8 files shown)


Ladicek:
Draft for now, because of the 3.2.0 target. Other than that, this should be ready.

core/src/main/java/org/jboss/jandex/EmptyIndex.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Collections;
+import java.util.Set;
+
+/**
+ * Empty index; that is, an index that doesn't contain any class. All methods return either
+ * an empty collection, or {@code null}.
+ *
+ * @since 3.2.0
+ */
+public final class EmptyIndex implements IndexView {
```
mkouba, line 13: Ehm, what is this for? I mean it's not even used internally, is it? And what's the benefit of this optimized empty version?  Right now, there are other ways to obtain an empty index such as `Index.of(Collections.emptyList())`. Shouldn't we use the `EmptyIndex` in similar scenarios? Maybe we should add something like `Index.empty()`?

Ladicek: So there's a difference between empty `Index` (which is fairly heavy-weight) and empty `IndexView` (which is super light-weight).

I don't have a use case for it myself, but apparently Hibernate Search people do: https://github.com/hibernate/hibernate-search/blob/main/util/common/src/main/java/org/hibernate/search/util/common/jar/impl/JandexUtils.java#L51 And it feels right to have a dedicated implementation directly in Jandex.

core/src/main/java/org/jboss/jandex/StackedIndex.java
```diff
@@ -0,0 +1,326 @@
...
+public final class StackedIndex implements IndexView {
+    // note that the top-most index comes first (reverse order compared to the `create()` methods)
+    private final IndexView[] stack;
+
+    private StackedIndex(IndexView[] stack) {
+        this.stack = stack;
+    }
+
+    /**
+     * Creates a stacked index from given {@code indexes}. The first element of the list is the bottom-most index
```
mkouba, line 29: This means that when you build the stack you need to add the indexes with "lower priority" first which is IMO not exactly intuitive and practical. I wonder if we should offer some more flexible way of building the stack.

core/src/main/java/org/jboss/jandex/EmptyIndex.java
mkouba: If so, then we should probably promote the API a bit more, i.e. mention/link it in the `IndexView`, maybe even add `IndexView.empty()` etc.

core/src/main/java/org/jboss/jandex/StackedIndex.java
Ladicek: I actually had a `push(IndexView)` method implemented (that would return a new stacked index with the given argument on top of the stack), but I felt like a solution in search of a problem, so I ultimately removed it. It should be easy to re-add.

core/src/main/java/org/jboss/jandex/EmptyIndex.java
Ladicek: I did contemplate about a `public static IndexView empty()` method on `IndexView` vs. a `public class EmptyIndex` and didn't feel strongly about either. I can switch to the `static` method for sure.

core/src/main/java/org/jboss/jandex/StackedIndex.java
Ladicek: Added `StackedIndex.pushIndex()`.

core/src/main/java/org/jboss/jandex/EmptyIndex.java
Ladicek: Added `IndexView.empty()`.

mkouba: I think that it's ok to have both.

Ladicek: Yeah, agree, I didn't know why I didn't make `EmptyIndex.INSTANCE` public. Fixed.

PR #347, opened 2023-12-11
Reduce hash collisions when interning type variable references

### LINKED ISSUE #322: Indexing takes over 100x longer on big file

Ladicek:
Fixes #322


Ladicek:
Draft for now, because of the 3.2.0 target. Other than that, this should be ready.

PR #354, opened 2024-03-04
improve `Index[View].getKnownUsers()`

Ladicek:
See also https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/hibernate.20reactive/near/424410591

### DIFF
--- a/core/src/main/java/org/jboss/jandex/Indexer.java
+++ b/core/src/main/java/org/jboss/jandex/Indexer.java
@@ -35,6 +35,7 @@ import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -417,7 +418,7 @@ public final class Indexer {
     private Map<DotName, List<ClassInfo>> implementors;
     private Map<DotName, ClassInfo> classes;
     private Map<DotName, ModuleInfo> modules;
-    private Map<DotName, List<ClassInfo>> users;
+    private Map<DotName, Set<ClassInfo>> users; // must be a linked set for reproducibility
     private NameTable names;
     private GenericSignatureParser signatureParser;
     private final TmpObjects tmpObjects = new TmpObjects();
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jboss/jandex/Indexer.java
```diff
@@ -2557,7 +2622,11 @@ public Index complete() {
         propagateTypeVariables();
 
         try {
-            return new Index(masterAnnotations, subclasses, subinterfaces, implementors, classes, modules, users);
+            Map<DotName, List<ClassInfo>> userLists = new HashMap<>();
+            for (Map.Entry<DotName, Set<ClassInfo>> entry : users.entrySet()) {
+                userLists.put(entry.getKey(), new ArrayList<>(entry.getValue()));
```
mkouba, line 2627: I wonder if `List#copyOf()` could be a little bit more efficient here?

Ladicek: Possibly, but Jandex is still on Java 8 :-)

mkouba: Ah, sorry. I forgot about it :facepalm:

mkouba (approved):
Looks good but I'm not 100% sure it's a good idea to target a bugfix release...

Ladicek:
I often add miniature features in micro releases, but you're probably right -- this should go into 3.2.0.

I'll change the milestone and make this a draft for now.

FroMage (approved):
I suppose this is more general, and more useful than just appearing in the constant pool, although I suppose that in the first two use-cases we've had (calling static methods on `Log` and calling any method/field on Panache entities) it's now going to be possible to get "users" that "only" declare our types (`Log` or Panache entities) somewhere in their signatures without calling any method on them, which may be "false positives" in these cases.

In the case of Panache, we'll register a class visitor/transformer too many, but it won't do anything since it won't find any field access to transform. The chances of classes having signature references to Panache entities without using their fields is plausible.

In the case of Quarkus `Log` it may be that we define an extra static logger for a class that will go unused, but the chances of that happening is pretty unlikely.

Ladicek:
Agree. Thanks!

PR #398, opened 2024-05-31
Improve reproducibility by sorting the Interned Indexes.

chirino:


chirino:
Right now I'm running into a weird issue where the FieldInternal pool seems to contain two equal objects by the time we create the iterator at: https://github.com/smallrye/jandex/pull/398/files#diff-9e59701eb76c28bada79887bf73b356179c79202ee57587dc186b232038f5f99R504

Odd thing is that I have an assertion in the compare method to double check that https://github.com/smallrye/jandex/pull/398/files#diff-02062034d4ac60db32d95b1ef268dcd1aca668998e955eb4f2f70f832d972081R129 the objects being compare really are equal.

core/src/main/java/org/jboss/jandex/AnnotationTarget.java
```diff
@@ -504,4 +504,30 @@ default Collection<AnnotationInstance> declaredAnnotationsWithRepeatable(Class<?
...
+        }
+
+        int v = o1.kind().compareTo(o2.kind());
+        if (v != 0) {
+            return v;
+        }
+
+        // TODO: is this good enough? Should we compare the actual objects?
+        v = o1.toString().compareTo(o2.toString());
+        if (v != 0) {
```
chirino, line 527: Yep.. we likely would need to make this better.

chirino:
Figured it out.. I needed to implement comparable in some sub classes.

PR #409, opened 2024-07-02
Added indexVersion parameter to the jandex-jar goal

martinhickson:
If you're happy with this change to add the indexVersion parameter to the jandex-jar goal (it's already present in the jandex goal), then it would be appreciated if you could apply this change to the base repository either via this pull request, or independently.


Ladicek (approved):

Ladicek:
Thanks!

PR #419, opened 2024-08-08
add ClassInfo.canonicalConstructor()

Ladicek:


core/src/main/java/org/jboss/jandex/ClassInfo.java
```diff
@@ -746,6 +746,50 @@ public final List<MethodInfo> constructors() {
...
+     * <p>
+     * Note that this method has the same limitations as {@link #unsortedRecordComponents()}. That is,
+     * at most 256 record components may be present, and an assumption is made that bytecode order
+     * of record components corresponds to the declaration order.
+     *
+     * @return the canonical constructor of this record, or {@code null} if this class is not a record
+     */
+    public MethodInfo canonicalConstructor() {
+        if (!isRecord()) {
+            return null;
```
mkouba, line 760: Isn't it better to throw an exception here?

Ladicek: In Jandex, it is customary to not throw but return a sentinel value in such cases. For example, `ClassInfo.recordComponents()` doesn't throw on non-records, it just returns an empty list.

We do throw an exception at the end of this method, but that's only for truly exceptional situations.

mkouba: Ok, `java.lang.Class.getRecordComponents()` also returns `null`. Sometimes we don't retun `null`, e.g. empty list in case of `ClassInfo#enumConstants()` but that's fine. TBH there should be probably `RecordInfo` and `EnumInfo` instead but :shrug:.

Ladicek: Yeah, that would also require `AnnotationInfo` and `InterfaceInfo` for completeness :-)

gsmet: Is there any way to make it clear from the name that this only works for records?

I'm not sure `canonical` will be clear enough for us laymen.

mkouba: That is a good point. It should be `canonicalRecordConstructor()`!

Ladicek: It seems to me that `canonicalConstructor()` is enough, but `canonicalRecordConstructor()` is good too. Will rename.

gsmet: Laymen thank you :).

Ladicek: https://github.com/smallrye/jandex/pull/421

FroMage: Yeah, agreed. So far, only records have canonical constructors. People might wonder what a non-record canonical constructor is, but it doesn't exist. So +1 on the changed name.

PR #420, opened 2024-08-08
replace ClassInfo.unsorted* with *InDeclarationOrder

Ladicek:


FroMage:
This is nice, because completion will list those methods when completing for `fields` and `methods`. Much better than before !

PR #488, opened 2025-01-31
Add JandexReflection.loadType()

Ladicek:
The `loadType()` method returns an implementation of `java.lang.reflect.Type`.
If possible, it is compatible with the JDK implementation, in the sense that
`equals()` and `hashCode()` work as expected. Unfortunately, the JDK makes
that impossible if the type contains type variables.


geoand:
Super useful!

We can certainly use that RESTEasy Reactive!

PR #498, opened 2025-02-13
Fix detection of Object type in AnnotationOverlayImpl#annotation

### LINKED ISSUE #497: AnnotationOverlayImpl#annotation does not detect Object properly when searching for inherited annotations
I originally encountered this with Quarkus and CDI BCE extensions observing `Object` type, not sure how common it would be otherwise.

The code in `AnnotationOverlayImpl#annotation` attempts to search for inherited annotations in super classes of given `Declaration` even if it is already `Object` type. This can lead to unexpected calls to `IndexView#getClassByName()` with `null` as parameter.

The code I am describing is here - https://github.com/smallrye/jandex/blob/main/core/src/main/java/org/jboss/jandex/AnnotationOverlayImpl.java#L135

I'll send a PR momentarily.

manovotn:
Fixes #497 

@Ladicek I did not yet figure out how to write a test for this in just Jandex (I have original reproducer in Quarkus), hints are welcome :)


Ladicek:
Thanks! Testing should be fairly simple, just create an index with `Object` (`Index.of(Object.class)`), an annotation overlay on top of it, and attempt to get annotations.

Ladicek:
Also, the same issue occurs in other methods. Could you please fix them too?

manovotn:
Sure, I'll get back to it tomorrow

manovotn:
I've pushed a change to other methods having similar issue and created a test using custom `IndexView`.
@Ladicek please re-review when you get the chance :)

Ladicek (approved):

Ladicek:
Thanks!

PR #527, opened 2025-05-13
fix ClassInfo.simpleName()

### LINKED ISSUE #526: ClassInfo#simpleName() behaves inconsistently for nested/top-level classes with a dollar sign in name
If there is a class like (I know `$` in the name is discouraged but still...):
```java
public class Top$Notch {}
```
Then the following test will fail:
```java
public class ClassInfoSimpleNameTest {

    @Test
    public void test() throws IOException {
        Index index = Index.of(Foo$Bar.class, Top$Notch.class);
        ClassInfo foo = index.getClassByName(Foo$Bar.class);
        assertEquals("Foo$Bar", foo.simpleName());
        ClassInfo top = index.getClassByName(Top$Notch.class);
        assertEquals("Top$Notch", top.simpleName());
    }

    public static class Foo$Bar {
    }

}
```
with `ClassInfoSimpleNameTest.test:19 expected: <Top$Notch> but was: <Notch>`.

In other words, for nested `Foo$Bar` the `ClassInfo#simpleName()` returns the correct simple name but for `Top$Notch` the `DotName#local()` is called and it returns `Notch`. I wonder if `DotName#withoutPackagePrefix()` should not be used for top-level classes instead?

Ladicek:
The `simpleName()` method used to return a wrong result in case of top-level classes with dollar `$` in their names. For example, for the following class

```java
public class Top$Level {
}
```

the `ClassInfo.simpleName()` method used to return `Level`. This commit fixes that by only calling `name.local()` in case the `name` is componentized (which it always is, so just to be sure) and is _not_ marked inner. If it is marked inner, the `simpleName()` method calls `name.withoutPackagePrefix()`.

It would be possible to call `withoutPackagePrefix()` always, but that method allocates in case of componentized names, which we want to avoid if possible.

Fixes #526

### DIFF
--- a/core/src/main/java/org/jboss/jandex/ClassInfo.java
+++ b/core/src/main/java/org/jboss/jandex/ClassInfo.java
@@ -1200,7 +1200,14 @@ public final class ClassInfo implements Declaration, Descriptor, GenericSignatur
      * @return the simple name of a top-level, member or local class, or {@code null} if this is an anonymous class
      */
     public String simpleName() {
-        return nestingInfo != null ? nestingInfo.simpleName : name.local();
+        // it would be enough to always call `name.withoutPackagePrefix()`, but we want to avoid needless allocations
+        if (nestingInfo != null) {
+            return nestingInfo.simpleName;
+        } else if (name.isComponentized() && !name.isInner()) {
+            return name.local();
+        } else {
+            return name.withoutPackagePrefix();
+        }
     }
 
     String nestingSimpleName() {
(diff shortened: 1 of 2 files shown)


mkouba:
> It would be possible to call `withoutPackagePrefix()` always, but that method allocates in case of componentized names, which we want to avoid if possible.

I think that we can't use the `withoutPackagePrefix()` for nested classes because if you declare a nested class `Bar` on a top-level class `Foo` then the simple name should be `Bar` but `DotName#withoutPackagePrefix()` would return `Foo$Bar`, I guess...

Ladicek:
Hmm, yeah, that's true. My bad. The comment is wrong. The code is right :-)

PR #570, opened 2025-08-21
Make index serializations reproducible

### LINKED ISSUE #568: Reproducible Jandex indexes
We're looking into having reproducible builds.
I've figured out the two Jandex indexes generated for the exact same `.class` files produce semantically equal results, but the serialized binary indexes are different.

Looking a bit deeper, the type-table, type-list-table and users are serialized in different orders.

I see that for example the name-table ensures stable (ordered) serialization using `org.jboss.jandex.IndexWriterV2#sortedNameTable`, but other tables that are based on `org.jboss.jandex.IndexWriterV2.ReferenceTable` do not.

I try to come up with a patch, and it seems it's going in the right direction. (My initial attempt didn't consider the dependency of array-types to the component types.)

snazy:
Jandex indexes produced with the current implementation are not reproducible, in the sense that multiple serializations of the same index produce the same binary representation. Having reproducible binary representations is important for reproducible builds, in order to verify a build result by rerunning a build.

The only reason that the serialized representation is not reproducible is due to letting `Set`s and `Map` dictate the order in which elements like classes, modules and usages are serialized.

The changes to make index serialization reproducible are luckily not very intrusive. It is mainly implemented via a new `IndexWriteView` class, which extracts the necessary collections from an `Index` and provides sorted `List`s used during serialization. Another part is the processing order of the "class elements" member-classes, permitted-subclasses and annotation-entries, which are now "guarded" with a sort operation.

No index format changes are necessary, all indexes serialized with `IndexWriterV2` benefit from this change. `IndexWriterV1` (for the not even anymore documented index format versions) has not been changed.

Fixes #568

### DIFF
--- a/core/src/test/java/org/jboss/jandex/test/ReproducibleSerializedIndexTest.java
+++ b/core/src/test/java/org/jboss/jandex/test/ReproducibleSerializedIndexTest.java
@@ -28,8 +28,6 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.RepeatedTest;
 
-import test.BridgeMethods;
-
 public class ReproducibleSerializedIndexTest {
     static List<String> testDataClassNames;
     static Map<String, byte[]> classData;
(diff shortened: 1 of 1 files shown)


Ladicek:
I wonder if you could put together a small project that isn't reproducible, because Jandex _should_ be reproducible since 3.1.0 I think (https://smallrye.io/blog/jandex-3-1-0/). We even have a test (`maven-plugin/src/it/reproducible`), so if you can make that test fail, that would be even better.

Ladicek:
I don't think the tests with reverse order or random order are valid. If you feed input to Jandex in different order, it should be expected that you'll get different indices. However, the test with identical order to original indeed fails often, so Jandex isn't really reproducible. Type annotations are a frequent source of troubles :-)

snazy:
@Ladicek could you let CI run on this PR?

test-data/pom.xml
```diff
@@ -12,7 +12,7 @@
     <name>Jandex: Test Data</name>
 
     <properties>
-        <java.version>17</java.version>
+        <java.version>8</java.version>
```
Ladicek, line 15: I'm not reviewing this PR properly yet, but this is not gonna work -- the `test-data` artifact uses Java 8+ features, like modules or records, so it cannot be built for Java 8.

PR #577, opened 2025-09-02
cache `toString()` for `DotName`s starting with `java.`

Ladicek:


core/src/main/java/org/jboss/jandex/DotName.java
```diff
@@ -326,14 +353,27 @@ public String toString() {
      */
     public String toString(char delim) {
         if (componentized) {
-            StringBuilder builder = new StringBuilder();
+            if (delim == '.' && startsWithJava()) {
+                return JAVA_STRINGS.computeIfAbsent(this, COMPONENTIZED_TO_STRING);
```
gsmet, line 357: A long time ago, @Sanne recommended me to do a simple `get()` before calling `computeIfAbsent()` in hot paths as it would perform better. We ended up using this pattern when we were pretty sure most of the lookups would be successful:

```
String name =  JAVA_STRINGS.get(this);
if (name != null) {
   return name;
}
return JAVA_STRINGS.computeIfAbsent(this, COMPONENTIZED_TO_STRING);
```

Now, I'm not sure if it's still true with modern Java but back in the days, it was performing better.

And given this thing will be a hotspot, I think it makes sense.

core/src/main/java/org/jboss/jandex/DotName.java
```diff
@@ -326,14 +353,27 @@ public String toString() {
...
     }
 
+    private int stringLength() {
+        int length = local.length();
+        DotName prefix = this.prefix;
+        while (prefix != null) {
+            length += 1 + prefix.local.length();
+            prefix = prefix.prefix;
+        }
+        return length;
```
gsmet, line 374: FWIW, in my initial patch, I used the same pattern as for `buildString()` so maybe you want to drop recursion from there too as you apparently weren't fan of it?

PR #579, opened 2025-09-02
add `ClassDesc` caching to `Jandex2Gizmo.classDescOf()`

### LINKED ISSUE #575: Jandex2Gizmo#classDescOf() should use the ClassDesc cache
I haven't created a patch for it as it will require some changes in Gizmo (as the cache is currently `Class` based) but I think it would be of value to share the `Util#constantCache` cache and make sure we can also use it in `Jandex2Gizmo#classDescOf()`.

I suspect we will have to make the cache key a `String` instead of a `Class` and avoid using `ClassValue`. Naively, I don't think it would be a problem as the `ClassDesc` is class loader-agnostic but let's wait for @dmlloyd's take on this.

Image

Ladicek:
Resolves #575

### DIFF
--- a/gizmo2/src/main/java/org/jboss/jandex/gizmo2/Jandex2Gizmo.java
+++ b/gizmo2/src/main/java/org/jboss/jandex/gizmo2/Jandex2Gizmo.java
@@ -10,6 +10,8 @@ import java.lang.constant.ConstantDescs;
 import java.lang.constant.MethodTypeDesc;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 
 import org.jboss.jandex.AnnotationInstance;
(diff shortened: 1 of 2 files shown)


gizmo2/src/main/java/org/jboss/jandex/gizmo2/Jandex2Gizmo.java
```diff
@@ -48,29 +50,33 @@
...
-                case "int" -> ConstantDescs.CD_int;
-                case "long" -> ConstantDescs.CD_long;
-                case "float" -> ConstantDescs.CD_float;
-                case "double" -> ConstantDescs.CD_double;
-                case "char" -> ConstantDescs.CD_char;
-                default -> ofClassOrArray(local);
-            };
-        }
-        return ofClassOrArray(name.toString());
+        return CLASS_DESC_CACHE.computeIfAbsent(name, nameParam -> {
```
gsmet, line 62: Probably worth using the same pattern as explained in: https://github.com/smallrye/jandex/pull/577#discussion_r2315516211 , especially since we parallelize class generation.

dmlloyd:
No, currently the only cache in Gizmo uses `ClassValue` in such a way that when a class is unloaded, its corresponding cache entries are also unloaded. However this trick is very specific to keying by `Class` *and* having the value be a class which is present in the JDK.

The plan for a more general cache at present is to have an unbounded cache plus a "clear the whole cache" method. It's not a great plan, but it's better than nothing.

PR #584, opened 2025-09-09
Add Smallrye icon for IntelliJ project list

snazy:


Ladicek:
We unfortunately don't have a logo for Jandex itself. Using the SmallRye logo makes sense, I guess, but we should take it from https://design.jboss.org/smallrye/

Ladicek (approved):

PR #589, opened 2025-09-15
Ensure that unannotated known class types are singletons

snazy:
In particular, this commit prevents "changing" an unbounded wildcard `?` to `? extends java.lang.Object`. This was caused by a new `ClassType` instance being created for `java.lang.Object`, instead of reusing the singleton instance.


Ladicek:
I've noticed this commit in the previous PR and didn't pay much attention, but I'd like to ask here: what issue does this fix? The PR looks reasonable, but still.

snazy:
This seemed much easier than an instance-equals at the call sites.

Ladicek:
Ah so accidental discovery. OK, I think this makes sense; `? extends java.lang.Object` with no annotation on the bound is the same as `?`.

snazy:
Or see it as a micro-heap-pressure-optimization ;)

PR #618, opened 2025-11-17
Use updated `ObjectOps` API

dmlloyd:
Draft until the next Gizmo2 release. Note that CI is expected to fail because of the `SNAPSHOT` dependency. https://github.com/quarkusio/gizmo/pull/557 will make this necessary.

Use the updated `ObjectOps` API and descriptor constants.

The `invokeInstance` method is planned to be removed after the next release because it necessarily causes the creation of a new method descriptor and type on every call, which adds up.


Ladicek:
Changes LGTM, please let me know when you want to merge this.

PR #707, opened 2026-08-07
Deduplicate CompositeIndex results by class name (#174)

### LINKED ISSUE #174: Get rid of Set<ClassInfo> usages
I've found 22 usages of `Set<ClassInfo>` directly in jandex classes `org.jboss.jandex.Index` and `org.jboss.jandex.CompositeIndex`. Given the fact that `ClassInfo` does not implement equals/hashCode it would make sense to refactor these classes, esp. the `CompositeIndex` where multiple `ClassInfo` instances representing the same java class may appear.

basteez:
Fixes #174.

## Problem

`ClassInfo` defines no `equals`/`hashCode`, so the `HashSet<ClassInfo>` accumulators in `Index` and `CompositeIndex` deduplicate by object identity. In an **overlapping `CompositeIndex`** (the same class indexed in more than one member index) this let the same fully-qualified class name appear **once per index** in the results of:

- `getKnownDirectSubclasses` / `getAllKnownSubclasses`
- `getKnownDirectSubinterfaces` / `getAllKnownSubinterfaces`
- `getKnownDirectImplementations`
- `getKnownDirectImplementors` / `getAllKnownImplementors`

This was inconsistent with `getKnownUsers`, `getClassesInPackage`, and `getAllKnownSubinterfaces`' worklist, which already deduplicate by `clazz.name()`.

## Change

Replaced the internal `Set<ClassInfo>` accumulators/parameters/locals in `Index` and `CompositeIndex` with name-keyed `Map<DotName, ClassInfo>` (`LinkedHashMap` for deterministic order; `putIfAbsent`/`containsKey` = first-writer-wins, matching `getClassByName`, which returns the first index's match). The traversal/visited-tracking logic (`processedClasses`, worklists) is unchanged.

- **Public API is preserved.** Declared return types are identical — the methods declared `Set<ClassInfo>` still return a `Set` (materialized from the deduplicated map values at the boundary); `Collection<ClassInfo>` returners still return a `Collection`. No `IndexView` signature, no `equals`/`hashCode` on `ClassInfo`, no on-disk index-format change. `Indexer` is untouched.
- Base source set stays Java 8 compatible.

## Behavior changes (please review)

1. **Deduplication by name** for the methods above on an overlapping `CompositeIndex`. This is the substance of #174. Since @Ladicek noted he wasn't fully convinced the identity semantics were safe to drop, I've kept this deliberate and easy to review: `IndexNavigationTest.testOverlappingCompositeIndex` previously asserted the duplicate-preserving behavior and is updated here to assert the deduplicated results. Happy to adjust the direction based on your preference.
2. A few `getAll*` methods that previously returned a mutable set now return an **unmodifiable** collection, consistent with the sibling `getKnownDirect*` methods that already did.

## Tests

- Added `CompositeTestCase.testCompositeDeduplicatesByClassName`: two indexes each contributing a *distinct* `ClassInfo` instance for the same FQCN; asserts the composite returns exactly one entry per name across the affected accessors.
- Updated `IndexNavigationTest.testOverlappingCompositeIndex` to expect deduplicated-by-name results.
- Full `core` build is green (302 tests) on JDK 21, formatter + impsort clean.

### DIFF
--- a/core/src/main/java/org/jboss/jandex/CompositeIndex.java
+++ b/core/src/main/java/org/jboss/jandex/CompositeIndex.java
@@ -26,7 +26,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 
@@ -104,14 +107,16 @@ public class CompositeIndex implements IndexView {
      */
     @Override
     public Set<ClassInfo> getKnownDirectSubclasses(final DotName className) {
-        final Set<ClassInfo> allKnown = new HashSet<ClassInfo>();
+        final Map<DotName, ClassInfo> allKnown = new LinkedHashMap<DotName, ClassInfo>();
         for (IndexView index : indexes) {
             final Collection<ClassInfo> list = index.getKnownDirectSubclasses(className);
             if (list != null) {
-                allKnown.addAll(list);
+                for (ClassInfo clazz : list) {
+                    allKnown.putIfAbsent(clazz.name(), clazz);
+                }
             }
         }
-        return Collections.unmodifiableSet(allKnown);
+        return Collections.unmodifiableSet(new LinkedHashSet<ClassInfo>(allKnown.values()));
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Ladicek:
Handling of duplicate classes is not documented at all in `CompositeIndex`, so technically anything goes, but changing it feels like huge behavioral breakage. But the changes here that ensure unmodifiable collections are returned are good, I'd accept those.

